PR #13, opened 2021-10-17
Update development setup documentation

elsehow:
This PR documents an `.env.local` gotcha. It also removes instructions on chain setup, directing developers to look at the cw-dao repo instead. This latter change prevents us from having to sync instructions as development flows change, possibly leaving developers with stale instructions in the process.


ben2x4:
Can we keep the note to make sure docker is installed?

PR #20, opened 2021-10-18
Add unit tests for validation proposal JSON

elsehow:
This PR improves validation for proposal JSON.

- It adds unit testing for proposal validation code.
- It adds a step to `pages/proposals/create.tsx` that valides the input is valid JSON.
- It adds more helpful error messages ("JSON is invalid" vs. "JSON contains invalid RPC messages.").

### DIFF
--- a/pages/proposals/create.tsx
+++ b/pages/proposals/create.tsx
@@ -63,17 +63,15 @@ const ProposalCreate: NextPage = () => {
     // check that proposal is valid json
     try {
       json = JSON.parse(jsonClone);
-    } catch {
-      setLoading(false)
-      setError('Proposal is not valid JSON.')
-      return
-    }
-    if (jsonClone) {
       if (!isValidJson(json)) {
         setLoading(false)
         setError('Proposal JSON is not a list of valid RPC messages.')
         return
       }
+    } catch {
+      setLoading(false)
+      setError('Proposal is not valid JSON.')
+      return
     }
 
     const msg = {
--- a/util/isValidJson.ts
+++ b/util/isValidJson.ts
@@ -18,6 +18,7 @@
  *
  */
 
+const COSMOS_TYPE_KEY = '@type'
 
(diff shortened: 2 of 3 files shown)


JakeHartnell (approved):

util/isValidJson.ts
```diff
@@ -0,0 +1,57 @@
...
+  // if it's parseable, check that it's valid
+  try {
+    // valid messages look like this:
+    //    { bank: { send: {"@type": ... } } }
+    // we'll go two levels deep to take the inner object
+    let inner = innerValue(innerValue(json))
+    // if it has a valid, string-y @type,
+    let typeDef: string = inner['@type']
+    // it's fine for now
+    return typeDef.length > 0
```
gavindoughtie, line 40: code golf:

`return !!typeDef?.length;`

pages/proposals/create.tsx
```diff
@@ -78,11 +59,19 @@ const ProposalCreate: NextPage = () => {
...
+    try {
       json = JSON.parse(jsonClone);
-      if (!validateJsonMsg(json)) {
+    } catch {
+      setLoading(false)
+      setError('Proposal is not valid JSON.')
+      return
+    }
+    if (jsonClone) {
+      if (!isValidJson(json)) {
```
gavindoughtie, line 72: you can move this check to right after the parse:
```
try {
      json = JSON.parse(jsonClone);
      if (!isValidJson(json)) {
         throw 'Invalid';
       }
    } catch {
      setLoading(false)
      setError('Proposal is not valid JSON.')
      return
    }
```

gavindoughtie (approved):

util/isValidJson.ts
elsehow: i realized this test misses some cases;  new commit has a less golfy but more thorough test. (and a few new test cases).

PR #25, opened 2021-10-21
Frontend for messages

gavindoughtie:
Haven't been able to do much during the week but thought I'd get this code into main. Mostly semicolons.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,5 +1,5 @@
 {
-  "semi": false,
+  "semi": true,
   "trailingComma": "es5",
   "singleQuote": true,
   "tabWidth": 2,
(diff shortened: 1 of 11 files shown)


.prettierrc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "semi": false,
+  "semi": true,
```
ebaker, line 2: hmm... this is the only thing i'm not sure about. i've been coding JS without semi's for a few years and don't see much value in using them - i think they unnecessarily clutter the code base. the only argument i have in favor of semi's is that they're needed in rust. but i'm not sure i think one code base's semantics should be based on another semantics of a completely different language. 

i think if we do want to change to semi's it should be done in a separate PR (as it's showing a lot of changes in the diff and currently CI is broken).

i made a script in package.json `yarn run prettier:write` which will prettier all the fils in the repo. i think we should change this back to `false` and run `prettier:write`. then we can handle changing to semi's in a separate PR once we're on the same page about this change.

PR #31, opened 2021-11-01
Full proposal json

gavindoughtie:
Editable JSON for the entire proposal, with editable messages

### DIFF
--- a/components/JsonEditor.tsx
+++ b/components/JsonEditor.tsx
@@ -5,10 +5,12 @@ import 'jsoneditor/dist/jsoneditor.css'
 export default function JsonEditor({
   json,
   onJsonChange,
+  onJsonError = (e: any) => {},
   ...props
 }: {
   json: any
   onJsonChange: (arg0: any) => void
+  onJsonError: (e: any) => void
 }) {
   const containerRef: MutableRefObject<any | null> = useRef(null)
   const editorRef: MutableRefObject<any | null> = useRef(null)
@@ -23,13 +25,11 @@ export default function JsonEditor({
           try {
             const json = editorRef.current.get()
             onJsonChange(json)
-          } catch (e) {
-            console.error(e)
+          } catch (e: any) {
+            onJsonError(e)
           }
         },
       } as any
-      // options['modes'] = ['view', 'tree', 'form', 'text']
-      // options['mode'] = 'text'
       const editor = new BaseJsonEditor(containerRef.current, options)
       editor.setMode('code')
       editor.set(json)
@@ -43,5 +43,10 @@ export default function JsonEditor({
     }
   }, [containerRef, json, props, onJsonChange])
 
-  return <div style={{ minHeight: '200px' }} ref={containerRef} />
+  return (
+    <div
+      className="min-h-full"
+      ref={containerRef}
+    />
+  )
 }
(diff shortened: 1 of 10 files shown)


JakeHartnell (approved):
LGTM

components/JsonEditor.tsx
```diff
@@ -43,5 +43,10 @@ export default function JsonEditor({
     }
   }, [containerRef, json, props, onJsonChange])
 
-  return <div style={{ minHeight: '200px' }} ref={containerRef} />
+  return (
+    <div
+      className="min-h-full"
```
ebaker, line 48: nice tailwind 👍

models/proposal/proposalForMessage.ts
```diff
@@ -0,0 +1,40 @@
+import { EmptyProposal, Proposal } from './proposal'
+import { ProposalReducer } from './proposalReducer'
+
+export function proposalForMessage(json: any): Proposal {
+  let proposal: Proposal = { ...EmptyProposal }
```
ebaker, line 5: can this be `const`?

components/ProposalEditor.tsx
```diff
@@ -236,6 +240,65 @@ export default function ProposalEditor({
...
+  let mainEditor
+  if (editProposalJson) {
+    mainEditor = (
+      <RawEditor
+        json={messageForProposal(proposal)}
+        onChange={handleJsonChanged}
+      ></RawEditor>
+    )
+  } else {
+    mainEditor = (
```
ebaker, line 261: i still feel like this should get broken out into a separate component. let's leave it here for now and continue iterating, just wanted to make a note

models/proposal/proposalForMessage.ts
```diff
@@ -0,0 +1,40 @@
...
+  return updateProposalForMessage(proposal, json)
+}
+
+export function updateProposalForMessage(
+  proposal: Proposal,
+  json: any
+): Proposal {
+  let valid = true
```
ebaker, line 13: can valid be false? don't see a case now, just checking

pages/proposals/[proposalId].tsx
```diff
@@ -176,6 +176,15 @@ const Proposal: NextPage = () => {
       <LineAlert msg={initialMessage} variant={initialMessageStatus} />
     ) : null
 
+  let proposalMessageContent = null
+  if (proposal?.msgs?.length) {
```
Comment (ebaker) [inline pages/proposals/[proposalId].tsx:180]: i feel like this would be more clear as a ternary, what do you think? also seems like another place it might make sense to create a component outside of this render, but we can leave it here for now for quicker iteration

ebaker (approved):
nice work! 👍 

overall looks good 🙂  i think there's some places we could break things out into components to trim the logic in the render functions, but we can break the components out later as we iterate

models/proposal/proposalForMessage.ts
gavindoughtie: Even better it can be inlined

models/proposal/proposalForMessage.ts
gavindoughtie: I was thinking we'd have some draft proposal updates where the message isn't valid but deleting for now

pages/proposals/[proposalId].tsx
Comment (gavindoughtie) [reply pages/proposals/[proposalId].tsx:180]: Definitely moved to a ternary in this case. I do like moving things to sub-components but I wait for "this is feeling gross" because there's some overhead in every separate component file you have to maintain. I think of the proposal component as a "shell" around several other components that reflect the proposal's contents. What were you thinking was a logical sub-group of those?

PR #32, opened 2021-11-02
quick fix for editor jank

gavindoughtie:
There's an issue with the JSON editor implementation that means we can't currently re-render it inside a component without losing the user's context.


ebaker (approved):
lgtm 👍

PR #35, opened 2021-11-03
New create ux

JakeHartnell:
Did this on the plane! Added a markdown editor too.

### DIFF
--- a/components/CustomEditor.tsx
+++ b/components/CustomEditor.tsx
@@ -5,6 +5,7 @@ import {
 import { ProposalAction } from 'models/proposal/proposalActions'
 import React, { useState } from 'react'
 import JSONInput from 'react-json-editor-ajrm'
+// @ts-ignore
 import locale from 'react-json-editor-ajrm/locale/en'
 
 export default function CustomEditor({
@@ -43,11 +44,11 @@ export default function CustomEditor({
   }
 
   // Handles values from react-json-editor-ajrm
-  function handleMessage(msg) {
+  function handleMessage(msg: any) {
     if (!msg.error) {
-      setError(undefined)
       setMessage(msg.json)
       updateCustom()
+      setError(undefined)
     } else {
       setError(msg.error)
     }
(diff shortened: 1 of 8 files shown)


components/ProposalCard.tsx
```diff
@@ -87,15 +87,13 @@ export default function ProposalCard({
   status,
   expires_at,
 }: ProposalCardProps) {
-  const expiresAtDateTime = new Date(expires_at / 1000000).toLocaleString()
+  //// TODO fix, expires_at time
```
gavindoughtie, line 90: I use this convention for todos: `TODO(user.name): message` which is nice for grepping with a lot of contributors.

gavindoughtie (approved):
Code LGTM.

This is probably better for quick demos in Lisbon, though I'm concerned about the drivability of the UI as things get longer/larger. Also sad to see the fancy JSON editor gone. Do you think we should bring it back for v1?

components/CustomEditor.tsx
```diff
@@ -0,0 +1,71 @@
...
+}: {
+  dispatch: (action: ProposalAction) => void
+  customMsg: MessageMapEntry
+}) {
+  const [message, setMessage] = useState(JSON.stringify(customMsg.message))
+  const [error, setError] = useState(undefined)
+
+  function updateCustom() {
+    try {
+      const id = customMsg?.id ?? ''
```
gavindoughtie, line 22: You want this to be `message?.id` don't you?

components/CustomEditor.tsx
```diff
@@ -0,0 +1,71 @@
...
+    } catch (err) {
+      console.error(err)
+    }
+  }
+
+  // Handles values from react-json-editor-ajrm
+  function handleMessage(msg) {
+    if (!msg.error) {
+      setError(undefined)
+      setMessage(msg.json)
```
gavindoughtie, line 49: I think this is in the incorrect sequence and makes the proposal update lag behind

JakeHartnell:
> Code LGTM.
> 
> This is probably better for quick demos in Lisbon, though I'm concerned about the drivability of the UI as things get longer/larger. Also sad to see the fancy JSON editor gone. Do you think we should bring it back for v1?

Can bring it back if we get it working nicely.

components/ProposalCard.tsx
JakeHartnell: Added a link to the new git issue I just created for this.

PR #41, opened 2021-11-08
Json editor fixes

gavindoughtie:
Video here: https://photos.app.goo.gl/FgxdDpEDPNuBr9cM7


JakeHartnell:
Looks like it's working now... feel the UX could be slightly better... but we can do that later. Is there a way we can do this without the save button?

Last thing, we need to run prettier so linting passes. You can install prettier in VS Code (or emacs) so it happens automatically.

gavindoughtie:
> Looks like it's working now... feel the UX could be slightly better... but we can do that later. Is there a way we can do this without the save button?
> 
> Last thing, we need to run prettier so linting passes. You can install prettier in VS Code (or emacs) so it happens automatically.

Done. Also was able to remove the save button. I dislike unstable UI, and the default error reporting in the control shifts the content around (hence my "status" area above it). Something to noodle around with, possibly will want to handle JSON editors as part of the reduced state and report it in the line that currently displays the message label ("CUSTOM json stuff")

JakeHartnell (approved):

PR #678, opened 2022-06-29
i18n: Ukrainian 🇺🇦

bmorphism:
image


packages/i18n/next-i18next.config.js
```diff
@@ -4,7 +4,7 @@ const path = require('path')
 const config = {
   i18n: {
     // Add your language ISO 639-1 code here.
-    locales: ['en', 'it'],
+    locales: ['en', 'it', 'pl', 'uk', 'zh'],
```
elsehow, line 7: Let's make this just `uk`, then should be good to go!

elsehow (changes requested):

elsehow (approved):

Issue #275, opened 2022-01-25
Re integrate proposal drafts


elsehow:
See https://github.com/DA0-DA0/docs/issues/30

Steps to reproduce:
- Make a proposal
- Press "Delete proposal draft"

Nothing happens, with no message in the console.

NoahSaso:
Completed in v2.

PR #748, opened 2022-07-20
[WIP] proposal drafts 

ebaker:
#### Issue
#275 

#### Screenshot

image

#### TODO
- refactor logic to better handle edge cases (navigating from draft to another drafts, prefill, & clear draftIds that don't exist)


NoahSaso:
Drafts implemented in v2. Thanks for your work on this @ebaker which inspired the latest implementation!

PR #754, opened 2022-07-21
i18n: Chinese 🇨🇳

bmorphism:
Xili (Sherry) has done this in a jiffy, I am helping with local testing and the right structure.

Please add to repo and reward w/ translation @xilimoocow

中国特色社会主义3


packages/i18n/next-i18next.config.js
```diff
@@ -4,7 +4,7 @@ const path = require('path')
 const config = {
   i18n: {
     // Add your language ISO 639-1 code here.
-    locales: ['en', 'it'],
+    locales: ['en', 'it', 'pl', 'uk', 'zh'],
```
elsehow, line 7: Can we add just the `zh` to this PR?

(I'm going to merge this first and get a second Mandarin-speaking reviewer to give it a pass).

elsehow (changes requested):

packages/i18n/next-i18next.config.js
JakeHartnell: IMO, let's get these all in and then try to get feedback from those who speak the language.

elsehow (approved):

PR #800, opened 2022-08-10
Support no quorum in update voting config action

### LINKED ISSUE #786: Detect existing voting config when update voting config proposal is created
https://discord.com/channels/895922260047720449/895922260655874080/1003709574303453294

Current 'update voting config' action does not detect existing voting config when creating a new proposal.

Specifically, if the DAO has quorum turned off, the proposal still loads an option to change quorum %age threshold. If possible, maybe we can consider an on/off toggle for the quorum action so that users have the option to leave it off, or turn it on. When "on" is selected, the %age threshold option can appear.

Furthermore, it might be worth allowing users to customize the voting configuration actions even further. That way the user can simply "remove" or "turn off" whichever settings are not relevant to their proposal. For example, if you only wanted to change the voting period from 1 day to 3 days, the other settings (thresholds, add/remove users, etc.) could be removed.

NoahSaso:
Resolves #786 

Adds support for enabled/disabled quorum to the `cw-proposal-single` proposal module adapter's UpdateVotingConfig action.
Screen Shot 2022-08-10 at 4 02 14 PM


ebaker (approved):
lgtm :+1:

PR #802, opened 2022-08-11
Fixed staking modal with proposal deposit

NoahSaso:
Added back button to stake all tokens except the proposal deposit, using the maximum proposal deposit from available proposal module adapters.

Also default to max (or max less proposal deposit if exists) when showing staking modal.

image

### DIFF
--- a/apps/dapp/pages/dao/[address]/index.tsx
+++ b/apps/dapp/pages/dao/[address]/index.tsx
@@ -5,7 +5,7 @@ import axios from 'axios'
 import { getAverageColor } from 'fast-average-color-node'
 import type { GetStaticPaths, GetStaticProps, NextPage } from 'next'
 import { useRouter } from 'next/router'
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import {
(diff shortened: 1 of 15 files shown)


apps/dapp/pages/dao/[address]/index.tsx
```diff
@@ -104,12 +132,31 @@ const InnerDAOHome = () => {
...
+          }).hooks.useDepositInfo
+      ),
+    [coreAddress, proposalModules]
+  )
+  const proposalModuleDepositInfos = useDepositInfoHooks
+    .map((useDepositInfo) =>
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      useDepositInfo?.()
+    )
+    .filter(Boolean) as CheckedDepositInfo[]
```
Comment (elsehow) [inline apps/dapp/pages/dao/[address]/index.tsx:158]: What's going on in this line? This has "magical chained method" energy. Perhaps a comment would help?

packages/voting-module-adapter/adapters/cw20-staked-balance-voting/components/Membership/index.tsx
```diff
@@ -83,6 +86,18 @@ const InnerMembership: FC<Omit<MembershipProps, 'primaryText'>> = ({
   const [showStakingMode, setShowStakingMode] = useState<StakingMode>()
   const stakingLoading = useRecoilValue(stakingLoadingAtom)
 
+  // Get max proposal deposit that needs the governance token.
```
elsehow, line 89: Let's make this comment more specific by telling the what and why of what we're doing. Something like:

```
// Of all the proposal modules this DAO uses, find the highest proposal deposit.
// We'll suggest that the user reserve that amount for proposals in our staking modal.
```

elsehow (changes requested):

elsehow (approved):

JakeHartnell (approved):

PR #809, opened 2022-08-15
Display loading in page titles before translations are ready

NoahSaso:
Right now, as translations load, the page title says `meta.title`. This replaces those temporary translations with `Loading...`.


ezekiiel (approved):
truly, i18n has to be the largest hack of all time. thanks for dealing with this stuff 🙏

JakeHartnell (approved):
Very nice.

PR #814, opened 2022-08-15
Improved onPlusMinus NumberInput API

NoahSaso:
The `onPlusMinus` prop of `NumberInput` is currently an array of functions that control what happens when you click the plus or minus sign. The functions are not named, which makes it easy to accidentally flip the order of the functions and cause the buttons to act contrary to the expected action.

This PR replaces that API with two props, `onMinus` and `onPlus`, to make it more readable and prevent bugs.

### DIFF
--- a/packages/ui/components/input/NumberInput.tsx
+++ b/packages/ui/components/input/NumberInput.tsx
@@ -15,9 +15,8 @@ interface NumberInputProps<FieldValues, FieldName extends Path<FieldValues>>
   register: UseFormRegister<FieldValues>
   validation?: Validate<FieldPathValue<FieldValues, FieldName>>[]
   error?: FieldError
-  defaultValue?: string
-  step?: string | number
-  onPlusMinus?: [() => void, () => void]
+  onMinus?: () => void
+  onPlus?: () => void
   containerClassName?: string
   sizing?: 'sm' | 'md' | 'auto'
   required?: boolean
@@ -38,9 +37,8 @@ export const NumberInput = <FieldValues, FieldName extends Path<FieldValues>>({
   register,
   error,
   validation,
-  defaultValue,
-  step,
-  onPlusMinus,
+  onMinus,
+  onPlus,
   disabled,
   sizing,
   className,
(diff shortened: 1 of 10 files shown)


packages/ui/components/input/NumberInput.tsx
```diff
@@ -54,73 +52,59 @@ export const NumberInput = <FieldValues, FieldName extends Path<FieldValues>>({
     {}
   )
 
-  const sharedProps = {
```
baoskee, line 57: Looks like there's more changes here than just `onPlus` and `onMinus`. Did this refactor change the view functionality at all or how it looks?

packages/voting-module-adapter/adapters/cw4-voting/actions/makeManageMembersAction/ManageMembersComponent.tsx
```diff
@@ -76,10 +76,10 @@ export const ManageMembersComponent: ActionComponent<ManageMembersOptions> = ({
                     disabled={!isCreating}
                     error={errors?.toAdd?.[index]?.weight}
                     fieldName={weightFieldName}
-                    onPlusMinus={[
-                      () => setValue(weightFieldName, Math.max(weight - 1, 0)),
-                      () => setValue(weightFieldName, weight + 1),
-                    ]}
+                    onMinus={() =>
```
baoskee, line 79: I think you mixed up the plus and minus functions here. The plus should be first and the minus second in `onPlusMinus`.

packages/ui/components/input/NumberInput.tsx
```diff
@@ -54,73 +52,59 @@ export const NumberInput = <FieldValues, FieldName extends Path<FieldValues>>({
...
 
-  return (
-    <input
-      className={clsx('body-text', _containerClassName)}
-      {...sharedProps}
-    />
+      <input
+        className={clsx(
+          'w-full text-right bg-transparent border-none outline-none ring-none body-text',
+          className
```
baoskee, line 97: Why was `_containerClassName` deleted here?

packages/ui/components/input/NumberInput.tsx
NoahSaso: Ah yeah, I should have explained. We had unnecessarily duplicated code, if there were plus/minus functions provided or not. I decided it was fine to use the same container-wrapped input if there are no plus/minus buttons since it looks the same and makes the code more readable.

packages/voting-module-adapter/adapters/cw4-voting/actions/makeManageMembersAction/ManageMembersComponent.tsx
NoahSaso: Great catch! This was the motivation for making this change. The changes should fix the issue, since the incorrect `onPlusMinus` array is removed here.

packages/ui/components/input/NumberInput.tsx
NoahSaso: See first conversation about refactor ^^

packages/voting-module-adapter/adapters/cw4-voting/actions/makeManageMembersAction/ManageMembersComponent.tsx
baoskee: I see. The `onPlusMinus` is incorrect.

baoskee (approved):
Everything makes sense. Although I don't understand the `_containerClassName` changes since I've never interacted with the input code.

PR #817, opened 2022-08-15
Hide IBC denom behind tooltip if the asset is unrecognized.

ezekiiel:
This fixes layout issues caused by IBC denoms that are too long.

Before:
image

After:
image

Hovering over the information icon shows the whole denom:
image

### DIFF
--- a/packages/ui/components/ContractView/BalanceIcon.tsx
+++ b/packages/ui/components/ContractView/BalanceIcon.tsx
@@ -19,3 +19,9 @@ export const BalanceIcon: FC<BalanceIconProps> = ({ iconURI }) => {
     ></div>
   )
 }
+
+export const UnknownAssetBalanceIcon = () => (
+  <div className="flex justify-center items-center w-4 h-4 text-black bg-disabled rounded-full">
+    ?
+  </div>
+)
--- a/packages/ui/components/ContractView/TreasuryView.tsx
+++ b/packages/ui/components/ContractView/TreasuryView.tsx
@@ -1,6 +1,7 @@
 import { FC } from 'react'
 import { useTranslation } from 'react-i18next'
 
+import { CopyToClipboard } from '@dao-dao/ui'
 import {
   NATIVE_DENOM,
   convertMicroDenomToDenomWithDecimals,
@@ -8,7 +9,6 @@ import {
   nativeTokenLogoURI,
 } from '@dao-dao/utils'
 
-import { TooltipIcon } from '../TooltipIcon'
 import { BalanceIcon, UnknownAssetBalanceIcon } from './BalanceIcon'
 import { BalanceListItem } from './BalanceListItem'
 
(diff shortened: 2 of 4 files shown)


packages/ui/components/ContractView/TreasuryView.tsx
```diff
@@ -44,6 +45,25 @@ export const TreasuryBalances: FC<TreasuryBalancesProps> = ({
...
+            <BalanceListItem key={symbol}>
+              <UnknownAssetBalanceIcon />
+              {convertMicroDenomToDenomWithDecimals(
+                amount,
+                decimals
+              ).toLocaleString(undefined, {
+                maximumFractionDigits: decimals,
+              })}{' '}
+              <span className="flex flex-row gap-1">
+                <p>Unknown Asset</p>
```
NoahSaso, line 61: Not sure why the linter passed. Would you kindly move this to the en translation (probably `info.unknownAsset`) :D

NoahSaso (changes requested):
Nice! Just need to create a translation key.

I was also thinking it would be nice for this string to be easily copyable. Is there a clean way to make tooltips copyable and indicate that to the user?

ezekiiel:
> @NoahSaso: it would be nice for this string to be easily copyable.

How about ditching the tooltip entirely and just doing a copy option? Have updated the PR to do this as I feel this is probably better UX.

image

elsehow (approved):
This looks great to me!

packages/ui/components/ContractView/BalanceIcon.tsx
```diff
@@ -19,3 +19,9 @@ export const BalanceIcon: FC<BalanceIconProps> = ({ iconURI }) => {
     ></div>
   )
 }
+
+export const UnknownAssetBalanceIcon = () => (
+  <div className="flex justify-center items-center w-4 h-4 text-black bg-disabled rounded-full">
+    ?
```
ebaker, line 25: sweet icon :slightly_smiling_face:

packages/ui/components/CopyToClipboard.tsx
```diff
@@ -28,19 +29,24 @@ interface CopyToClipboardProps {
...
   value,
   success = 'Copied to clipboard!',
   takeN,
   takeStartEnd,
+  className = 'font-mono text-xs',
 }) => {
   const [copied, setCopied] = useState(false)
 
   return (
     <button
```
ebaker, line 45: can we set the `title` on this to the un-concat address?> this way when you hover mouse you'll see full address. what do you think?

ebaker (approved):
nice and timely haha!

lgtm :+1:

packages/ui/components/CopyToClipboard.tsx
ezekiiel: Great idea!

NoahSaso (approved):

PR #819, opened 2022-08-17
Use autogenerated featured DAOs list.

ezekiiel:
This swaps us over to using the autogenerated featured DAOs list @ebaker recently built: https://github.com/DA0-DA0/web2.5-dao-index/pull/3

It also brings in the placeholder images from the v2 designs and uses those as fallbacks on the homepage for when a featured DAO doesn't have a valid image URL.


apps/dapp/components/home/FeaturedDAOsList.tsx
```diff
@@ -3,17 +3,20 @@
 
 import { PlusIcon, SparklesIcon } from '@heroicons/react/outline'
 import Link from 'next/link'
+import { FC } from 'react'
```
ebaker, line 6: is this the recommended use of `FC` @NoahSaso ?

apps/dapp/pages/home.tsx
```diff
@@ -7,27 +7,43 @@ import { serverSideTranslations } from '@dao-dao/i18n/serverSideTranslations'
...
 export default HomePage
 
-export const getStaticProps: GetStaticProps = async ({ locale }) => ({
-  props: {
-    ...(await serverSideTranslations(locale, ['translation'])),
-  },
-})
+export const getStaticProps: GetStaticProps = async ({ locale }) => {
+  const resp = await fetch(
+    'https://dao-stats.withoutdoing.com/mainnet/featured_daos.json'
```
ebaker, line 36: can we put this into env var?

apps/dapp/pages/home.tsx
```diff
@@ -7,27 +7,43 @@ import { serverSideTranslations } from '@dao-dao/i18n/serverSideTranslations'
...
+  )
+  // These are returned as a timeseries in the form [{time, value}, ...].
+  const featuredDaosOverTime = await resp.json()
```
ebaker, line 39: hmm.. i think this is fine for now, but we may want to rethink this since i don't think old data is ever shown on the page, so we are dl'ing extra data we never use. but i think it's fine for now 👍

apps/dapp/components/FeaturedCard.tsx
```diff
@@ -53,11 +64,8 @@ export const FeaturedCard = ({
 
       <div className="flex flex-col gap-1 mt-5 items-left">
         <p className="text-sm">
-          <Dao className="inline mr-2 mb-1 w-4" fill="currentColor" />${TVL} TVL
-        </p>
-        <p className="text-sm text-valid text-success">
-          <Votes className="inline mr-2 mb-1 h-5" fill="currentColor" />
-          {members} {t('info.members')}
+          <Dao className="inline mr-2 mb-1 w-4" fill="currentColor" />$
+          {Number(TVL).toLocaleString()} TVL
```
ebaker, line 68: do we want `toFixed(2)` or `toFixed(6)` to set # of decimals or does `toLocaleString()` handle this? do we also want to use `USDC` before tvl as i know we were concerned about showing USDC prices as USD

ebaker (approved):
thank you for integrating @ezekiiel !!

just a couple minor questions 🙂

apps/dapp/components/home/FeaturedDAOsList.tsx
```diff
@@ -3,17 +3,20 @@
...
-import { featuredDaos } from '@/util'
 
 import { FeaturedCard } from '../FeaturedCard'
+import { FeaturedDao } from '../splash'
 import { DaoCardContainer } from './DaoCardContainer'
 
-export const FeaturedDAOsList = () => {
+export const FeaturedDAOsList: FC<{ featuredDaos: FeaturedDao[] }> = ({
+  featuredDaos,
+}) => {
```
NoahSaso, line 19:
```suggestion
interface FeaturedDAOsListProps {
  featuredDaos: FeaturedDao[]
}

export const FeaturedDAOsList = ({ featuredDaos }: FeaturedDAOsListProps) => {
```

Using `FC` is now discouraged and will cause breaking changes in future react versions. It no longer implicitly includes `children` but is used that way in most React node until now. The recommended style is just
`export const Component = (props: PropsInterface) => {`

apps/dapp/components/splash/FeaturedDaos.tsx
```diff
@@ -10,11 +11,20 @@ import {
...
+  name: string
+  description: string
+  href: string
+  TVL: string
+  image: string
+}
+
+export const FeaturedDaos: FC<{ featuredDaos: FeaturedDao[] }> = ({
+  featuredDaos,
+}) => {
```
NoahSaso, line 27:
```suggestion
interface FeaturedDaosProps {
  featuredDaos: FeaturedDao[]
}

export const FeaturedDaos = ({ featuredDaos }: FeaturedDaosProps) => {
```

Same as above.

NoahSaso (changes requested):
Nice! Super minor changes requested.

apps/dapp/pages/home.tsx
ezekiiel: Agree. We can update the generation script later if needed. :)

apps/dapp/pages/home.tsx
ezekiiel: great idea aha

ezekiiel:
Added an environment variable for the URL and added it to the vercel deployments. Should be ready to go.

NoahSaso (approved):
gorgeous

PR #824, opened 2022-08-17
Fix undelegation balance validation

### LINKED ISSUE #806: Undelegating a native amount from a validator, fails if treasury balance is lower than input amount.
So basically you can't undelegate without the error validating a different way :(

Requirements:
* Get the native staked balance(s) for the DAO
* Change the value for undelegating to validate against the staked amount
* Optionally support a "MAX" button within the undelegate amount input. Making it easier to move funds

TrevorJTClarke:
This PR will fix #806


packages/actions/components/Stake.tsx
```diff
@@ -46,24 +46,48 @@ export const stakeActions: { type: StakeType; name: string }[] = [
...
+  const { register, watch, getValues, setError, clearErrors, setValue } =
+    useFormContext()
 
   const stakeType = watch(fieldNamePrefix + 'stakeType')
   const amount = watch(fieldNamePrefix + 'amount')
   const denom = watch(fieldNamePrefix + 'denom')
 
   const validatePossibleSpend = useCallback(
     (denom: string, amount: string): string | boolean => {
+      const delegateType = getValues(fieldNamePrefix + 'stakeType')
```
NoahSaso, line 69: This value is already retrieved on line 63 (`stakeType`), so you can just reuse it and add to this callback's dependency array below instead of `getValues`.

NoahSaso (changes requested):
Awesome! Thanks for adding this 🙏 

One minor change and should be good to go :)

packages/actions/components/Stake.tsx
TrevorJTClarke: Good call - pushed a fix

NoahSaso (approved):

PR #831, opened 2022-08-18
NFT Card Component

NoahSaso:
image
Screen Shot 2022-08-18 at 12 56 14 AM

On hover:
Screen Shot 2022-08-18 at 12 51 03 AM
Screen Shot 2022-08-18 at 12 56 28 AM

This also creates another component—`TooltipDisplay`—which is a new tooltip design since the `NftCard` component needs it:

Screen Shot 2022-08-18 at 12 53 17 AM
Screen Shot 2022-08-18 at 12 56 41 AM


ezekiiel (approved):
:shipit:

PR #835, opened 2022-08-19
created ProfileNotMemberCard component

ebaker:
#### Screenshots

image

image

#### Figma Node
https://www.figma.com/file/ZnQ4SMv8UUgKDZsR5YjVGH/DAO-DAO-2.0?node-id=94%3A14883


ezekiiel:
Looking good! For this time around we're really trying to be as pixel perfect to the designs as we can be. I think we've got a good system for this coming along with the latest color updates. Wrote a little wiki page about how to map designs to the css system here: https://github.com/DA0-DA0/dao-dao-ui/wiki/Implementing-designs-in-the-DAO-DAO-UI

packages/ui/components/profile/ProfileCardTemplate.tsx
```diff
@@ -0,0 +1,40 @@
...
+}: ProfileCardTemplateProps) => {
+  const { t } = useTranslation()
+
+  return (
+    <div className="flex flex-col justify-center items-center pt-10 rounded-lg border-2 border-border-primary">
+      <ProfileImage imgUrl={imgUrl} />
+      <div className="pt-6 title-text">{walletName}</div>
+      <div className="py-2 mb-4 font-mono caption-text">
+        {t('profile.est')}
+        {new Intl.DateTimeFormat('default', {
```
ebaker, line 30: @ezekiiel looked into adding this to i18n... but realized that's a js package. i don't think it makes sense to add ts to i18n package just for this. might make more sense to put this type of helper in utils once we can import it

packages/ui/styles/index.css
```diff
@@ -119,7 +119,7 @@ input[type='number']::-webkit-outer-spin-button {
     --text-secondary: rgba(var(--v2-dark), 0.7);
     --text-tertiary: rgba(var(--v2-dark), 0.5);
     --text-brand: rgba(var(--v2-brand), 0.95);
-    --text-interactive-disabled: rgba(var(--v2-light), 0.2);
+    --text-interactive-disabled: rgba(var(--v2-dark), 0.2);
```
ebaker, line 122: @NoahSaso feel free to double check this

packages/ui/components/profile/ProfileCardTemplate.tsx
NoahSaso: Yeah, agree that we should have a standard date formatter util

packages/ui/styles/index.css
NoahSaso: Hmm, I think both themes use `--v2-light` (looking at the css on the bottom right).

Screen Shot 2022-08-19 at 11 12 15 PM
Screen Shot 2022-08-19 at 11 12 43 PM

ezekiiel (commented):
Cool! Noticed two little things.

I also just put together a little websever for getting average colors for images: https://fac.withoutdoing.com/

Should let use do corner gradients.

packages/ui/styles/index.css
NoahSaso: Never mind, I think you're right :)

NoahSaso (approved):
Looks great!

PR #837, opened 2022-08-19
ProposalYourVote component

baoskee:
Screen Shot 2022-08-19 at 2 03 05 PM
storybook ftw

### DIFF
--- a/packages/ui/components/proposal/ProposalYourVote.tsx
+++ b/packages/ui/components/proposal/ProposalYourVote.tsx
@@ -12,7 +12,7 @@ export const ProposalYourVote = ({ variant }: YourVoteProps) => {
         variant === 'abstain'
           ? 'text-secondary bg-primary'
           : variant === 'pending'
-          ? 'border border-primary'
+          ? 'border-2 border-inactive'
           : variant === 'no'
           ? 'text-interactive-error bg-[#C73E5914]'
           : 'text-valid bg-[#39A69914]'
@@ -20,7 +20,7 @@ export const ProposalYourVote = ({ variant }: YourVoteProps) => {
     >
       {variant[0].toUpperCase() + variant.slice(1)}
       {variant === 'pending' && (
-        <div className="absolute top-0 right-0 w-3 h-3 bg-[#F53E86] rounded-full"></div>
+        <div className="absolute top-0 right-0 w-2 h-2 bg-[#F53E86] rounded-full"></div>
       )}
     </div>
   )
--- a/packages/ui/components/proposal/ProposalYourVote.tsx
+++ b/packages/ui/components/proposal/ProposalYourVote.tsx
@@ -10,7 +10,7 @@ export const ProposalYourVote = ({ variant }: YourVoteProps) => {
       className={clsx(
         'flex relative justify-center items-center py-1 px-4 w-20 text-sm font-medium rounded-full',
         variant === 'abstain'
-          ? 'text-text-secondary bg-primary'
+          ? 'text-text-secondary bg-background-secondary'
           : variant === 'pending'
           ? 'border-2 border-inactive'
           : variant === 'no'
--- a/packages/ui/components/proposal/ProposalYourVote.tsx
+++ b/packages/ui/components/proposal/ProposalYourVote.tsx
@@ -12,7 +12,7 @@ export const ProposalYourVote = ({ variant }: YourVoteProps) => {
         variant === 'abstain'
           ? 'text-text-secondary bg-background-secondary'
           : variant === 'pending'
-          ? 'border-2 border-inactive'
+          ? 'border-2 border-border-primary'
           : variant === 'no'
           ? 'text-text-interactive-error bg-[#C73E5914]'
           : 'text-text-valid bg-color-valid/[0.08]'
--- a/packages/ui/components/proposal/ProposalYourVote.tsx
+++ b/packages/ui/components/proposal/ProposalYourVote.tsx
@@ -14,7 +14,7 @@ export const ProposalYourVote = ({ variant }: YourVoteProps) => {
           : variant === 'pending'
           ? 'border-2 border-border-primary'
           : variant === 'no'
-          ? 'text-text-interactive-error bg-[#C73E5914]'
+          ? 'text-text-interactive-error bg-background-interactive-error'
           : 'text-text-valid bg-color-valid/[0.08]'
       )}
     >
--- a/packages/ui/components/proposal/ProposalYourVote.tsx
+++ b/packages/ui/components/proposal/ProposalYourVote.tsx
@@ -15,7 +15,7 @@ export const ProposalYourVote = ({ variant }: YourVoteProps) => {
           ? 'border-2 border-inactive'
           : variant === 'no'
           ? 'text-text-interactive-error bg-[#C73E5914]'
-          : 'text-text-valid bg-[#39A69914]'
+          : 'text-text-valid bg-color-valid/[0.08]'
       )}
     >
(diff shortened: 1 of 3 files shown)


NoahSaso:
Can you rebase on v2-redesign and change to merging into that one? You shouldn't need to add a color anymore :)

NoahSaso (changes requested):
Can you rebase on v2-redesign and change to merging into that one? You shouldn't need to add a color anymore :)

baoskee:
Fixed some problems 
Screen Shot 2022-08-22 at 4 49 40 PM

packages/ui/components/proposal/ProposalYourVote.tsx
```diff
@@ -1,5 +1,27 @@
...
+          : variant === 'pending'
+          ? 'border-2 border-inactive'
+          : variant === 'no'
+          ? 'text-interactive-error bg-[#C73E5914]'
+          : 'text-valid bg-[#39A69914]'
+      )}
+    >
+      {variant[0].toUpperCase() + variant.slice(1)}
+      {variant === 'pending' && (
+        <div className="absolute top-0 right-0 w-2 h-2 bg-[#F53E86] rounded-full"></div>
```
NoahSaso, line 23: Can we use the new color classes (e.g. `text-secondary` should now be `text-text-secondary` IIRC)? See @ezekiiel's amazing wiki page for more info: https://github.com/DA0-DA0/dao-dao-ui/wiki/Implementing-designs-in-the-DAO-DAO-UI

All these hardcoded colors are also probably now covered by the new color names :)

packages/ui/tailwind/colors.js
```diff
@@ -125,6 +125,7 @@ const textColor = {
   tertiary: generateColorClass('dark', 0.6),
   disabled: generateColorClass('dark', 0.4),
   accent: generateColorClass('accent', 1),
+  'interactive-error': '#C73E59F2',
```
NoahSaso, line 128: No need for this with new color system.

NoahSaso (changes requested):
Nice! Looking good.

Now is great time to update the colors to use the new system. More info here: https://github.com/DA0-DA0/dao-dao-ui/wiki/Implementing-designs-in-the-DAO-DAO-UI

packages/ui/components/proposal/ProposalYourVote.tsx
baoskee: Hey looks like some of the colors are not part of the theme so they do have to be hard-coded. If you click on the component and there is a theme color I use it, else I just hardcode it based on Francois's design.

packages/ui/components/proposal/ProposalYourVote.tsx
```diff
@@ -1,5 +1,27 @@
...
+  return (
+    <div
+      className={clsx(
+        'flex relative justify-center items-center py-1 px-4 w-20 text-sm font-medium rounded-full',
+        variant === 'abstain'
+          ? 'text-text-secondary bg-primary'
+          : variant === 'pending'
+          ? 'border-2 border-inactive'
+          : variant === 'no'
+          ? 'text-text-interactive-error bg-[#C73E5914]'
```
NoahSaso, line 17:
```suggestion
          ? 'text-text-interactive-error bg-background-interactive-error'
```

`#C73E59` corresponds to `rgb(199, 62, 89)`, which exists in the colors as `color-error`. Since tailwind lets you change the opacity of colors
([like this](https://tailwindcss.com/docs/background-color#changing-the-opacity)), this doesn't need to be hardcoded, so `bg-color-error/[0.08]` should work.

Alternatively, we can just use `bg-background-interactive-error` since it uses the error color with an opacity of `0.1`, which is negligibly different.

packages/ui/components/proposal/ProposalYourVote.tsx
```diff
@@ -1,5 +1,27 @@
...
+          : 'text-text-valid bg-[#39A69914]'
```
NoahSaso, line 18:
```suggestion
          : 'text-text-valid bg-color-valid/[0.08]'
```

`#39A699` corresponds to `rgb(57, 166, 153)`, which is `color-valid`, so `bg-color-valid/[0.08]` should work.

packages/ui/components/proposal/ProposalYourVote.tsx
```diff
@@ -1,5 +1,27 @@
...
+
+export const ProposalYourVote = ({ variant }: YourVoteProps) => {
+  return (
+    <div
+      className={clsx(
+        'flex relative justify-center items-center py-1 px-4 w-20 text-sm font-medium rounded-full',
+        variant === 'abstain'
+          ? 'text-text-secondary bg-primary'
+          : variant === 'pending'
+          ? 'border-2 border-inactive'
```
NoahSaso, line 15:
```suggestion
          ? 'border-2 border-border-primary'
```

using the new color system :D

packages/ui/components/proposal/ProposalYourVote.tsx
```diff
@@ -1,5 +1,27 @@
...
+  variant: 'abstain' | 'pending' | 'no' | 'yes'
+}
+
+export const ProposalYourVote = ({ variant }: YourVoteProps) => {
+  return (
+    <div
+      className={clsx(
+        'flex relative justify-center items-center py-1 px-4 w-20 text-sm font-medium rounded-full',
+        variant === 'abstain'
+          ? 'text-text-secondary bg-primary'
```
NoahSaso, line 13:
```suggestion
          ? 'text-text-secondary bg-background-secondary'
```

using the new color system :D

packages/ui/components/proposal/ProposalYourVote.tsx
```diff
@@ -1,5 +1,27 @@
...
+          : variant === 'pending'
+          ? 'border-2 border-inactive'
+          : variant === 'no'
+          ? 'text-text-interactive-error bg-[#C73E5914]'
+          : 'text-text-valid bg-[#39A69914]'
+      )}
+    >
+      {variant[0].toUpperCase() + variant.slice(1)}
+      {variant === 'pending' && (
+        <div className="absolute top-0 right-0 w-2 h-2 bg-[#F53E86] rounded-full"></div>
```
NoahSaso, line 23: this hardcoding is necessary because the color (`#F53E86` = `rgb(245, 62, 134)`) does not exist in `index.css`

packages/ui/stories/proposal/ProposalYourVote.stories.tsx
```diff
@@ -3,16 +3,20 @@ import { ComponentMeta, ComponentStory } from '@storybook/react'
 import { ProposalYourVote } from 'components/proposal/ProposalYourVote'
 
 export default {
-  title: 'DAO DAO UI V2 / proposal / ProposalYourVote',
+  title: 'DAO DAO UI v2 / proposal / ProposalYourVote',
```
NoahSaso, line 6:
```suggestion
  title: 'DAO DAO UI V2 / proposal / ProposalYourVote',
```

for consistency

NoahSaso (changes requested):
Few more color system changes :) then we'll be good to go

packages/ui/components/proposal/ProposalYourVote.tsx
baoskee: hey the change is pretty notice-able on dark mode. but we can fix this later. Thanks for the suggestions!

baoskee:
The suggestions are great hehe.

NoahSaso (approved):
nice :) thanks for dealing with all my color fix requests 😁  this looks great

PR #839, opened 2022-08-20
Add SegmentedControls component.

ezekiiel:
This adds a segmented controls component. I had initially schemed on doing some fancy animation work to make it slide, but in the designs it is laid out as a sequence of buttons. This is insanely clean to implement, and very familiar to users.

image
image

Might be nice to take an animated approach later, but for now this really feels like something where for 1/10th the work we can have 9.5/10ths as good of UX.


NoahSaso (approved):
noice :)

PR #845, opened 2022-08-24
Improve segmented control component

NoahSaso:

## Fixed stretching responsiveness and rounding when wrapping

Before:
Screen Shot 2022-08-24 at 1 35 14 AM

After:
Screen Shot 2022-08-24 at 1 37 00 AM

## Hides spacers on both sides when hovering to prevent the buttons from appearing like they're touching

Before:
Screen Shot 2022-08-24 at 1 34 55 AM

After:
Screen Shot 2022-08-24 at 1 37 34 AM
Screen Shot 2022-08-24 at 1 33 09 AM

## Added loader capability to tabs
Screen Shot 2022-08-24 at 1 34 28 AM

### DIFF
--- a/packages/ui/components/SegmentedControls.tsx
+++ b/packages/ui/components/SegmentedControls.tsx
@@ -1,10 +1,12 @@
 import clsx from 'clsx'
+import { useState } from 'react'
 
 import { Button } from './Button'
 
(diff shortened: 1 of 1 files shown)


ezekiiel (approved):
Thanks! Great catch.

PR #846, opened 2022-08-24
ProposalStatus component

NoahSaso:
image


packages/ui/components/proposal/ProposalStatus.tsx
```diff
@@ -1,5 +1,52 @@
-export interface ProposalStatusProps {}
+import clsx from 'clsx'
+import { ComponentType, SVGProps } from 'react'
 
-export const ProposalStatus = ({}: ProposalStatusProps) => {
-  return <p>placeholder</p>
+import { Approved, Open, Rejected } from '@dao-dao/icons'
+
+export enum EProposalStatus {
```
elsehow, line 6: Is it a convention to start Enum names with `E`? Separate question: is this convention linted?

packages/ui/components/proposal/ProposalStatus.tsx
```diff
@@ -1,5 +1,52 @@
...
+}
+
+export const ProposalStatus = ({ status }: ProposalStatusProps) => {
+  const { Icon, iconClassName, textClassName } = ProposalStatusMap[status]
+
+  return (
+    <div className="flex flex-row gap-2 items-center link-text">
+      <Icon className={clsx('w-[19px] h-[19px]', iconClassName)} />
+      <p className={textClassName}>
+        {status[0].toUpperCase() + status.slice(1)}
```
elsehow, line 23: This slicing is a little too magic. What's happening here?

packages/ui/stories/proposal/ProposalStatus.stories.tsx
```diff
@@ -7,12 +7,15 @@ export default {
   component: ProposalStatus,
 } as ComponentMeta<typeof ProposalStatus>
 
-const Template: ComponentStory<typeof ProposalStatus> = (args) => (
-  <ProposalStatus {...args} />
+const Template: ComponentStory<typeof ProposalStatus> = () => (
+  <div className="flex flex-row gap-12 items-center">
+    <ProposalStatus status="open" />
+    <ProposalStatus status="rejected" />
+    <ProposalStatus status="approved" />
```
elsehow, line 14: Any way we can use our enum for these? Conversely, any reason we might not want to use our enum for these?

elsehow (approved):
Looks great! A few little questions and comments.

packages/ui/components/proposal/ProposalStatus.tsx
NoahSaso: Not a convention, just wasn't sure what else to call it 😁

packages/ui/components/proposal/ProposalStatus.tsx
NoahSaso: Converting to titlecase basically. Capitalizes first letter followed by the rest, and since the enum values are lowercase, the rest is lowercase.

packages/ui/stories/proposal/ProposalStatus.stories.tsx
NoahSaso: This is a bit more readable, and the typescript `${enum}` type I used in the component lets you pass the enum or the value.

packages/ui/components/proposal/ProposalStatus.tsx
elsehow: Let's call it `ProposalStatusEnum`. Took me a sec to figure out what `E` meant.

packages/ui/components/proposal/ProposalStatus.tsx
elsehow: I think a `const convertToTitleCase = (str) => ...` could go a long way here.

packages/ui/stories/proposal/ProposalStatus.stories.tsx
elsehow: Got it. Why not add a comment that says this? Good to explain what's going to use that enum and how - in this case, the typechecker's use of that enum is implicit, but a comment would make it explicit.

packages/ui/components/proposal/ProposalStatus.tsx
NoahSaso: Agreed, this should be a util function, but we can't use those right now because storybook gets mad :( maybe we can add a TODO

packages/ui/stories/proposal/ProposalStatus.stories.tsx
NoahSaso: I'd rather use the enum than add a comment, so I'll change it!

PR #849, opened 2022-08-25
`ProfileHomeCard` component

NoahSaso:
Screen Shot 2022-08-24 at 9 11 30 PM

No proposals in inbox:

Screen Shot 2022-08-24 at 9 12 08 PM


packages/ui/components/profile/ProfileNotMemberCard.tsx
```diff
@@ -26,10 +27,11 @@ export const ProfileNotMemberCard = ({
 
   return (
     <ProfileCardWrapper
-      daoName={daoName}
       established={established}
       imgUrl={profileImgUrl}
-      isMember={false}
+      underHeaderComponent={
```
ebaker, line 32: nice 🙂

packages/ui/stories/profile/ProfileNotMemberCard.stories.tsx
```diff
@@ -8,7 +8,7 @@ export default {
 } as ComponentMeta<typeof ProfileNotMemberCard>
 
 const Template: ComponentStory<typeof ProfileNotMemberCard> = (args) => (
-  <div className="max-w-[320px]">
+  <div className="max-w-xs">
```
ebaker, line 11: sweet 👍

packages/ui/components/profile/ProfileHomeCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+          <div className="flex flex-col items-stretch text-center">
+            <Payments className="self-center mb-4 w-5 h-4 text-center text-icon-secondary" />
+            <p className="mb-1 secondary-text">{t('title.holdings')}</p>
+            <p className="font-mono break-words title-text">
+              {unstakedBalance.toLocaleString(undefined, {
+                // eslint-disable-next-line i18next/no-literal-string
+                notation: 'compact',
+                maximumFractionDigits: 2,
+              })}{' '}
+              ${tokenSymbol}
```
ebaker, line 52: i created https://github.com/DA0-DA0/dao-dao-ui/blob/03a7e31cc80c76bfede3bbdf190bf20a76aece10/packages/i18n/locales/en/translation.json#L199 lmk whether u think that's overkill in this sitation

ebaker (approved):
nice!! 🙂 

lgtm 👍

packages/ui/components/profile/ProfileHomeCard.tsx
NoahSaso: The important part of this is the `compact` and `maximumFractionDigits` that control its formatting. I don't know if the i18n library lets us automatically do compact notation (suffixing with K, M, B, or T automatically as the number increases), but if it is I'm down to make a new translation key :)

This is a very common pattern and either should be a translation key or a util function. I'm kinda thinking a util function because nothing about this needs to be translated. wdyt?

PR #856, opened 2022-09-02
Fix legacy page not loading due to missing block height

NoahSaso:
https://legacy.daodao.zone/multisig/juno1qv9rtaf0jm33ft5ruhunzmj2wl39v062q82ykxg4x43k0g997anqec4wmj

is crashing because the RPC can no longer load the block height information. This PR allows it to silently fail and return undefined instead of throwing an error.


ezekiiel (approved):
nice!

JakeHartnell (approved):

elsehow (approved):

PR #863, opened 2022-09-09
Authz and validator actions

JakeHartnell:
Adds actions for Authz:
- MsgExec: execute message on behalf of another account
- MsgGrant: grant another account to permission to preform certain TXs on the DAO's behalf
- MsgRevoke: remove previously granted authorization

Adds actions for working with validators:
- Create validator
- Edit validator
- Claim validator commission

Encoding these stargate types is a bit messy, perhaps @pyramation can come to the rescue?


JakeHartnell:
Will resubmit against v2.

PR #866, opened 2022-09-14
Fix redelegation balance validation

blue-note:


packages/actions/components/Stake.tsx
```diff
@@ -65,8 +65,11 @@ export const StakeComponent: ActionComponent<StakeOptions> = ({
...
     (denom: string, amount: string): string | boolean => {
-      // Logic for undelegating
-      if (stakeType === StakeType.Undelegate) {
+      // Logic for undelegating or redelegating
+      if (
+        stakeType === StakeType.Undelegate ||
+        stakeType === StakeType.Redelegate
+      ) {
         const humanReadableAmount = convertMicroDenomToDenomWithDecimals(
           nativeDelegatedBalance.amount,
```
NoahSaso, line 74: Is this `nativeDelegatedBalance` the total they have staked across all validators? We should probably use `StargateClient`'s `getDelegation(delegatorAddress: string, validatorAddress: string): Promise<Coin | null>` instead for a more precise validation, though that is not as urgent. Maybe add an issue to make the redelegation more precise for the validator address entered?

NoahSaso (approved):
Sweet!

JakeHartnell (approved):

packages/actions/components/Stake.tsx
blue-note: sure!

PR #871, opened 2022-09-19
Lower PROP_LOAD_LIMIT to prevent out-of-gas with 20+ proposals.

ezekiiel:
We can not definitively set this so that an out-of-gas error will never occur, but we can lower when we run into things. DAO that was failing: <https://daodao.zone/dao/juno1luw9kspq5dwrqrxgkvfwt443ue99umdaqmm57afg7ms5y0rkz3dsfeudxp>

### DIFF
--- a/packages/common/components/ProposalList.tsx
+++ b/packages/common/components/ProposalList.tsx
@@ -23,7 +23,7 @@ import { useDaoInfoContext } from './DaoPageWrapper'
 import { SuspenseLoader } from './SuspenseLoader'
 
 // Max = 30
-const PROP_LOAD_LIMIT = 30
+const PROP_LOAD_LIMIT = 25
 
 export interface ProposalListProps {
   proposalCreateUrl: string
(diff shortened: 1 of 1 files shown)


NoahSaso (approved):

packages/common/components/ProposalList.tsx
```diff
@@ -22,8 +22,9 @@ import {
 import { useDaoInfoContext } from './DaoPageWrapper'
 import { SuspenseLoader } from './SuspenseLoader'
 
-// Max = 30
-const PROP_LOAD_LIMIT = 30
+// Contracts enforce a max of 30, though this is on the edge for DAOs with
```
elsehow, line 25: Let's add a wee bit more context at the top of this comment here - "The number of proposals a DAO will load on the frontend." Rest of the comment is great. 

Would be nice to double click on what it means for 30 to be "on the edge." Do you mean, DAOs with big proposals will  out-of-gas if their proposals are exceptionally big? How big is too big? Can we _ever_ pick some number where we'll never see an out of gas error?

elsehow (approved):
This looks great. Added a comment about comments, as usual (:

JakeHartnell (approved):

PR #881, opened 2022-10-07
V2 redesign plus authz and validators

JakeHartnell:
Adds actions for Authz:
- MsgExec: execute message on behalf of another account
- MsgGrant: grant another account to permission to preform certain TXs on the DAO's behalf
- MsgRevoke: remove previously granted authorization

Adds actions for working with validators:
- Create validator
- Edit validator
- Claim validator commission

Todo:
- [ ] Unjail validator

PR #883, opened 2022-10-10
Add unjail validator action based on #881

tougerthao:

I added the unjail validator action.


packages/actions/actions/UnjailValidator.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+const useTransformToCosmos: UseTransformToCosmos<UnjailValidatorData> = () =>
+  useCallback(
+    (data: UnjailValidatorData) =>
+      makeStargateMessage({
+        stargate: {
+          type_url: data.type_url,
+          value: JSON.parse(data.value),
+        },
+      }),
```
JakeHartnell, line 42: So this is correct, but we need to also encode the `/cosmos.slashing/v1beta1.MsgUnjail` message in `makeStargateMessage`! If you look at that function you'll see similar examples for other messages.

JakeHartnell (approved):
Awesome!

PR #892, opened 2022-10-20
Add an empty state for the proposal inbox

sagarsaija:
Add the NoContent box for an empty proposal inbox:
Screen Shot 2022-10-19 at 5 56 10 PM

### DIFF
--- a/apps/dapp/pages/inbox.tsx
+++ b/apps/dapp/pages/inbox.tsx
@@ -1,6 +1,7 @@
 // GNU AFFERO GENERAL PUBLIC LICENSE Version 3. Copyright (C) 2022 DAO DAO Contributors.
 // See the "LICENSE" file in the root directory of this package for more copyright information.
 
+import { WhereToVoteOutlined } from '@mui/icons-material'
 import { useWallet } from '@noahsaso/cosmodal'
 import { GetStaticProps, NextPage } from 'next'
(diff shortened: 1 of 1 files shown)


NoahSaso:
There actually already exists an empty state that I think provides more clarity. It links to the pinned DAOs it checked for proposals and shows that there is nothing for each one, while also allowing you to visit their DAO page by clicking on them.

Screen Shot 2022-10-26 at 12 54 54 PM

What do you think about this?

NoahSaso (changes requested):
Apologies, I misunderstood what was going on here. This is a great idea!

Though I think the empty state for pinned proposals is more reflective of what's happening on the inbox page:
Screen Shot 2022-11-01 at 4 40 01 PM

How would you feel about using the same code from here:
https://github.com/DA0-DA0/dao-dao-ui/blob/83570b1a8ca93d096430a29dd63856e17e4345b5/packages/stateless/components/dao/PinnedDaos.tsx#L47-L55
and putting it in the stateless Inbox page here:
https://github.com/DA0-DA0/dao-dao-ui/blob/83570b1a8ca93d096430a29dd63856e17e4345b5/packages/stateless/pages/Inbox.tsx#L97

This ensures that it will only display the NoContent component once it finishes loading, because the array is empty when they are still loading. Thanks again for thinking about this.

PR #904, opened 2022-11-08
Overhauled docs.

NoahSaso:
Added `README.md` files to various adapters and systems that needed explanation, and added code layout documentation to each existing package README.

To review this, I'd recommend reading through the root README and jumping between them. They should all link together, or at least to folders containing README files.


elsehow (approved):
This is great!

PR #907, opened 2022-11-09
Added i18n diff script.

NoahSaso:
Created script to output the diff between translation files to generate an output like [this](https://gist.github.com/NoahSaso/80ee2886f81b0c19bfc2024cad26847f) which makes it easier for translators to see the changes.

This was used to compare the old V1 translations file with the new V2 translations file in an effort to ease the updating process.

I also preemptively inserted the label for the multiple choice proposal module adapter since that will be necessary shortly.


elsehow (approved):
This is a great idea. LGTM. Just make sure to document this process for translators, perhaps in the i18n readme.

PR #910, opened 2022-11-13
Fixed select input style and Stake action responsiveness

NoahSaso:
The Stake action was not very responsiveness, elements overflowing on small screens. This fixes that:

Screenshot 2022-11-13 at 12 00 53 PM
Screenshot 2022-11-13 at 12 01 13 PM

The select input was not highlighting on hover or anything like the other clickable input elements (like buttons), and also the dropdown arrow was weirdly positioned because of its default browser appearance. This hides that arrow and uses a new one:

Screenshot 2022-11-13 at 12 02 30 PM
Screenshot 2022-11-13 at 12 02 36 PM
Screenshot 2022-11-13 at 12 02 43 PM

I also found a recently introduced bug where the select input would not register the `onChange` handler when not registered in a form. This is now fixed.

### DIFF
--- a/packages/stateful/actions/components/Stake.tsx
+++ b/packages/stateful/actions/components/Stake.tsx
@@ -233,9 +233,10 @@ export const StakeComponent: ActionComponent<StakeOptions, StakeData> = ({
       onRemove={onRemove}
       title={t('title.stake')}
     >
-      <div className="flex flex-row gap-2">
+      <div className="flex flex-col gap-2 xs:flex-row">
         {/* Choose type of stake operation. */}
         <SelectInput
+          containerClassName="shrink-0"
           defaultValue={stakeActions[0].type}
           disabled={!isCreating}
           error={errors?.stakeType}
@@ -265,7 +266,7 @@ export const StakeComponent: ActionComponent<StakeOptions, StakeData> = ({
 
         {/* Choose source validator. */}
         <ValidatorPicker
-          displayClassName="grow"
+          displayClassName="grow min-w-0"
           nativeDecimals={NATIVE_DECIMALS}
           nativeDenom={NATIVE_DENOM}
           onSelect={({ address }) =>
@@ -285,7 +286,7 @@ export const StakeComponent: ActionComponent<StakeOptions, StakeData> = ({
 
       {/* If not claiming (i.e. withdrawing reward), show amount input. */}
       {stakeType !== StakeType.WithdrawDelegatorReward && (
-        <div className="flex flex-row gap-2">
+        <div className="flex flex-col gap-2 xs:flex-row">
           <NumberInput
             containerClassName="grow"
             disabled={!isCreating}
--- a/packages/stateful/actions/components/ValidatorPicker.tsx
+++ b/packages/stateful/actions/components/ValidatorPicker.tsx
@@ -85,7 +85,7 @@ export const ValidatorPicker = ({
       Trigger={({ open, ...props }) => (
         <div className={clsx('flex', displayClassName)}>
           {selectedAddress ? (
-            <InputThemedText className="flex min-w-0 grow flex-row items-center justify-between gap-10">
+            <InputThemedText className="flex min-w-0 grow flex-row items-center justify-between gap-2">
               <CopyToClipboard
                 label={selectedValidator?.moniker}
                 tooltip={t('button.clickToCopyAddress')}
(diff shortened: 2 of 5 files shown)


packages/stateful/actions/components/Stake.tsx
```diff
@@ -285,7 +286,7 @@ export const StakeComponent: ActionComponent<StakeOptions, StakeData> = ({
 
       {/* If not claiming (i.e. withdrawing reward), show amount input. */}
       {stakeType !== StakeType.WithdrawDelegatorReward && (
-        <div className="flex flex-row gap-2">
+        <div className="flex flex-col gap-2 xs:flex-row">
```
onewhiskeypls, line 289: **thought**: not tested, but could be the same thought as above?

```suggestion
        <div className="flex flex-col gap-2">
```

packages/stateful/actions/components/Stake.tsx
```diff
@@ -265,7 +266,7 @@ export const StakeComponent: ActionComponent<StakeOptions, StakeData> = ({
 
         {/* Choose source validator. */}
         <ValidatorPicker
-          displayClassName="grow"
+          displayClassName="grow min-w-0"
```
onewhiskeypls, line 269: **question**: is this one of those weird styling things where it wants a min h/w of 0 so flex can function properly?

packages/stateful/proposal-module-adapter/adapters/CwdProposalSingle/daoCreation/QuorumVotingConfigItem.tsx
```diff
@@ -49,7 +49,7 @@ export const QuorumInput = ({
           )}
 
           <SelectInput
-            className={majority ? 'grow' : undefined}
+            containerClassName={majority ? 'grow' : undefined}
```
onewhiskeypls, line 52: **praise**: lgtm

packages/stateless/components/inputs/SelectInput.tsx
```diff
@@ -41,27 +47,41 @@ export const SelectInput = <
...
   return (
-    <select
-      className={clsx(
-        'secondary-text rounded-md bg-transparent py-3 px-4 text-text-body ring-1 transition placeholder:text-text-tertiary focus:outline-none focus:ring-2',
-        error
-          ? 'ring-border-interactive-error'
-          : 'ring-border-primary focus:ring-border-interactive-focus',
-        props.disabled && 'appearance-none',
-        className
+    <div className={clsx('relative', containerClassName)}>
```
onewhiskeypls, line 50: **thought**: my laziness is kicking in and i dont want to go through each class here lol, but a quick check of this and it looks good!

packages/stateful/actions/components/ValidatorPicker.tsx
```diff
@@ -85,7 +85,7 @@ export const ValidatorPicker = ({
       Trigger={({ open, ...props }) => (
         <div className={clsx('flex', displayClassName)}>
           {selectedAddress ? (
-            <InputThemedText className="flex min-w-0 grow flex-row items-center justify-between gap-10">
+            <InputThemedText className="flex min-w-0 grow flex-row items-center justify-between gap-2">
```
onewhiskeypls, line 88: **praise**: lgtm

onewhiskeypls (approved):
**testedACK**: lgtm. I tested using firefox's responsive design mode. There seemed to be one unnecessary class used twice `xs:flex-row`, please correct me if I'm wrong!  I also took a quick look at other actions and their responsiveness and it looks like we'll need to clean those up too.

- Mint
- Spend
- Update Proposal Submission Config
- Update Voting Config

packages/stateful/actions/components/Stake.tsx
```diff
@@ -233,9 +233,10 @@ export const StakeComponent: ActionComponent<StakeOptions, StakeData> = ({
       onRemove={onRemove}
       title={t('title.stake')}
     >
-      <div className="flex flex-row gap-2">
+      <div className="flex flex-col gap-2 xs:flex-row">
```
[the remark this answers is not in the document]
NoahSaso: This converts it from being stacked vertically to being stacked horizontally once the screen size is large enough. This is the main responsive change actually.

packages/stateful/actions/components/Stake.tsx
NoahSaso: Same as above.

packages/stateful/actions/components/Stake.tsx
NoahSaso: Yeah sorta. Flex by default set its children `min-width` properties to `auto`, which means they will be as large as their content inside, overflowing the container instead of truncating text internally. Setting `min-width` to `0` means they will not overflow their containers and the text truncates! yay

elsehow (approved):

PR #917, opened 2022-11-18
Improved error text when simulating proposal actions fails

NoahSaso:
Made proposal action simulation failure error more specific.

Also removed `not found` from the common errors since it was potentially omitting information by replacing all errors that contain `not found` with `Not found.`


0xekez:
feel like the ux of checking the console is pretty sub optimal. can we show this error on the toast or on the page somewhere?

NoahSaso:
> feel like the ux of checking the console is pretty sub optimal. can we show this error on the toast or on the page somewhere?

Agreed, I realized the error was being omitted because of the CommonError.NotFound enum that was too generally applying itself, so I removed it. The toast should show the full error message now, so I removed the browser console piece.

0xekez:
ah, probably "migrate method not found" or something. their contract likely just didn't export that.

0xekez (approved):
:turbo:

PR #919, opened 2022-11-21
Show tooltips on disabled button links

NoahSaso:
Improved ButtonLink and IconButtonLink so that their disabled states prevent navigation but still allows tooltips to be shown. Anchor (link) tags (`<a>`) do not support the `disabled` prop like buttons do, so they need special handling to disable them.

The previous method was disabling all pointer events on the element, and the new method (this PR) is to override the `href` with a `#` value, which navigates nowhere. The disabled styles still show, and the cursor does not display a pointer, so it's clear that the button is disabled. Tooltips cannot be shown when an element has pointer events disabled, so if we want to show messages explaining why button links are disabled, we need this change.

This is a better solution because tooltips can be shown and links can still be disabled.


elsehow (approved):
This is a tricky situation, but this seems like a good solution for it. Really appreciate you doing this kind of tech debt paydown.

PR #922, opened 2022-11-26
Clean up button/link APIs and make DaoImage load when clicked

NoahSaso:
This is a continuation of #919, which fixed button links so that tooltips can be shown even when disabled. After further interaction, I decided the UX of this change is suboptimal and could use minor changes.

Replacing disabled links with `#` makes some browsers jump to the top of the page when clicked, which is not great for a disabled button. I revert back to using `pointer-events: none` to disable all touch interactions on the link, and add a `div` container to `LinkWrapper` that receives tooltip functionality instead of the `a` link directly. The `ButtonLink` and `IconButtonLink` components use `LinkWrapper` under the hood, so this should take care of all links.

I also noticed some components that were not taking advantage of the global stateful navigation state, such as the inbox button in `ProfileHomeCard` and `DaoImage`'s parent DAO link circle on `DaoCard`. This PR adds in the loading state to these where it was missing.

This PR also synchronizes `ButtonLink` and `IconButtonLink` APIs a bit, which should be more similar than they currently are. Loading and disabled states were handled differently, which made the code quite unreadable. Now the abstractions are stronger.

This PR is a mess of nuance and context and likely useful but definitely non-trivial abstractions. I recommend playing with the UI and making sure buttons and links function as expected.


packages/stateful/components/ButtonLink.tsx
```diff
@@ -9,38 +8,30 @@ import {
...
+        loading={loading || navigating}
         onClick={(event) => {
           onClick?.(event)
           // If not on destination page, set navigating state. If already there
           // or href is invalid, do nothing.
-          if (router.asPath !== href && href && href !== '#') {
+          if (
+            router.asPath !== href &&
+            href &&
+            href !== '#' &&
```
elsehow, line 31: A comment could be good here.

packages/stateful/components/IconButtonLink.tsx
```diff
@@ -7,38 +6,37 @@ import { navigatingToHrefAtom } from '@dao-dao/state/recoil'
...
+        onClick={(event) => {
+          onClick?.(event)
+          // If not on destination page, set navigating state. If already there
+          // or href is invalid, do nothing.
+          if (
+            router.asPath !== href &&
+            href &&
+            href !== '#' &&
+            // Don't set navigating if remote.
+            !href.startsWith('http')
```
elsehow, line 31: Feels a little redundant - is there any way to make this more DRY?

elsehow (approved):

packages/stateful/components/ButtonLink.tsx
NoahSaso: No longer needed! Removing.

packages/stateful/components/IconButtonLink.tsx
NoahSaso: Great idea.

NoahSaso: Great idea.

PR #930, opened 2022-11-29
Token swap actions

### LINKED ISSUE #897: Token swap action
An action to instantiate a [token swap contract](https://github.com/DA0-DA0/dao-contracts/tree/main/contracts/external/cw-token-swap).

NoahSaso:
Resolves #897 

This PR creates two actions. One lets you create a new token swap or fund an existing token swap. The other lets you withdraw from an incomplete token swap.

Side effects of this PR:
- Redefined the profile abstraction to be a superset of wallet profiles and DAO profiles. Now the `ProfileDisplay` component can be used with DAOs and wallets, allowing seamless integration of either party into the token swap components. This will be useful later.
- Wrote an `objectMatchesStructure` utility that allows more readable nested object key presence validation.

# Perform token swap

## Create new
Screenshot 2022-12-06 at 1 16 45 PM

## Fund existing
image
(Screenshots below taken before explanation text seen above was added)
Screenshot 2022-11-29 at 4 30 19 PM
Screenshot 2022-11-29 at 4 29 04 PM
Screenshot 2022-11-29 at 4 29 30 PM

## Once created or chose existing
Screenshot 2022-11-29 at 4 25 26 PM
Screenshot 2022-11-29 at 4 26 58 PM

# Withdraw token swap
Screenshot 2022-11-29 at 6 06 18 PM
Screenshot 2022-11-29 at 6 07 06 PM


JakeHartnell (approved):
:shipit:

packages/i18n/locales/en/translation.json
```diff
@@ -159,17 +164,23 @@
     "noNftsSelected": "No NFTs selected.",
     "noValidatorFound": "No validator found.",
     "noVotingPower": "You have not given anyone voting power. Add some members to your DAO.",
+    "notATokenSwapContractAddress": "This contract is not a token swap contract.",
```
elsehow, line 167: This is so great.

packages/i18n/locales/en/translation.json
```diff
@@ -159,17 +164,23 @@
...
     "notCw20Address": "This contract address is not a CW20 token.",
     "notCw721Address": "This contract address is not a NFT (CW721).",
     "notEnoughForDeposit": "You do not have enough tokens in your balance to pay the proposal deposit",
     "notInActiveSet": "This validator is not in the active set. No rewards will be earned while this is the case.",
+    "notPartyInTokenSwap_dao": "The DAO is not a party in this token swap.",
+    "notPartyInTokenSwap_wallet": "You are not a party in this token swap.",
+    "notYetSentTokenSwap_dao": "The DAO has not yet sent its share of this token swap, so it cannot withdraw.",
```
elsehow, line 174: I find this a little confusing. Does it mean that you can't withdraw the counterparty's share (because that share isn't in there yet), or you can't withdraw your share (because the swap is still open)? Depending on which this means, I might suggest a different text.

A main question for me is: what happens if the counterparty never funds? Can you withdraw your share still, or no?

packages/i18n/locales/en/translation.json
```diff
@@ -159,17 +164,23 @@
...
+    "notYetSentTokenSwap_wallet": "You have not yet sent your share of this token swap, so you cannot withdraw.",
```
elsehow, line 175: Same note as above - you cannot withdraw _what_ exactly?

packages/i18n/locales/en/translation.json
```diff
@@ -159,17 +164,23 @@
...
     "nothingStaked": "Nothing staked.",
     "pageNotFound": "We couldn't find that page.",
     "pfpkStargazeReopenModal": "Failed to connect to Stargaze. Reopen the modal to try again.",
     "simulationFailedInvalidProposalActions": "Simulation failed. Verify your proposal actions are valid.",
     "stakeInsufficient": "The DAO has {{amount}} ${{tokenSymbol}} staked, which is insufficient.",
     "subDaoAlreadyExists": "SubDAO already exists.",
-    "treasuryInsufficientCannotStake": "The treasury only has {{amount}} ${{tokenSymbol}}, which is insufficient.",
+    "tokenSwapContractNotChosen": "You must instantiate a token swap contract or enter the address of an existing token swap before using this action.",
+    "treasuryInsufficient": "The treasury only has {{amount}} ${{tokenSymbol}}, which is insufficient.",
```
elsehow, line 183: > The treasury only has {{amount}} ${{tokenSymbol}}, less than the {{proposed_amount}} you propose here.

packages/i18n/locales/en/translation.json
```diff
@@ -241,6 +256,7 @@
     "tokenContractAddressTitle": "Token contract address",
     "tokenDefinition": "Token definition",
     "tokenInfo": "Token info",
+    "tokenSwapInstantiateInstructions": "Before the token swap can be initiated, you must instantiate the smart contract. This is a one-time step that must be completed prior to submitting this action. Use the form below to instantiate the contract. No funds will be transferred at this time.",
```
elsehow, line 259: This one took me a while to figure out. What about:

> You must initiate a token swap before you fund it. This initiation step describes how many funds each party will need to send for the swap to complete. No funds will be transferred at this time.

packages/i18n/locales/en/translation.json
```diff
@@ -633,6 +662,7 @@
     "wallet": "Wallet",
     "watchOut": "Watch out!",
     "when": "When",
+    "withdrawTokenSwap": "Withdraw Token Swap",
```
elsehow, line 665: Again, withdraw what? Your share, or the counterparty's share?

elsehow (approved):
This is great, and super code quality. I've left a few notes on the copy. One thing we may want to talk about is the word "withdraw." It can mean two things: get the counterparty's tokens (i.e., complete the token swap), or claw back your share (i.e., cancel the token swap). Which does it mean? And can the initiating party back out of the token swap? If so, we should indicate how. If not, we should make sure people understand that this is irreversible! Also, what happens if the counterparty never completes? When does the swap expire? We'll need to communicate that to users, too. This is all stuff we can discuss offline. Otherwise, this PR looks great.

PR #940, opened 2022-12-02
use pfpk address display if showing a disabled address input.

0xekez:
if we're rendering a disabled `AddressInput` display the addresses' pfpk profile.

if the address has a pfpk profile:

image

if the address has no pfpk profile:

image


NoahSaso:
@0xekez this PR is no longer relevant because of #941 right?

PR #941, opened 2022-12-03
Show profiles when displaying address inputs 🌞

0xekez:
while editing an address, shows a preview of the profile of the address. while viewing an address input which is disabled, a profile is shown.

for example, while viewing a proposal to manage multisig membership:

image

and while editing an address input:

image


NoahSaso (approved):
DAMN 🔥 HUGE

PR #949, opened 2022-12-09
Compensation cycle status fixes

NoahSaso:
image

Fixes compensation cycle showing 'Not submitted' when wallet disconnected (by hiding any status badge). Also improves the status text to be more verbose.

(cc @elsehow let me know if I missed any fixes)


elsehow (approved):
fire

PR #950, opened 2022-12-09
Rename to align with dao-contracts rename (cwd --> dao)

NoahSaso:
Fixed URLs pointing to dao-contracts and updated the adapters to support the new contract naming scheme. I also got rid of some environmental variables that were unnecessary relating to contract names.


elsehow (approved):
Listen, this runs and passes the tests/linter, so I am satisfied. I'm not going to pretend like I read every line of this PR. Thank you for doing this inglorious work @NoahSaso.

PR #951, opened 2022-12-09
Only show token card stake unstake and claim buttons for the native denom.

NoahSaso:
Currently, the TokenCard displays stake/unstake/claim actions for all tokens, instead of just the native token. These actions are currently only valid for the native token, so let's only show them for that token card.


elsehow (approved):
perfect

PR #954, opened 2022-12-13
Moved average color fetch to recoil selector to take advantage of cache

NoahSaso:
Average color was being fetched in `useEffect`s which triggered on every render, constantly asking for the same request. This moves that request into a Recoil selector so it caches based on input.

### DIFF
--- a/packages/state/recoil/selectors/misc.ts
+++ b/packages/state/recoil/selectors/misc.ts
@@ -1,0 +1,23 @@
+import { selectorFamily } from 'recoil'
+
+export const averageColorSelector = selectorFamily({
+  key: 'averageColor',
+  get: (url: string) => async () => {
+    if (!url || url.startsWith('http://localhost')) {
+      return undefined
+    }
+
+    try {
+      const response = await fetch(`https://fac.withoutdoing.com/${url}`)
+      const color = await response.text()
+      // Validate color format.
+      if (!color.startsWith('#')) {
+        return undefined
+      }
+
+      return color
+    } catch (err) {
+      console.error(err)
+    }
+  },
+})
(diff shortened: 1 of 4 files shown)


packages/state/recoil/selectors/misc.ts
```diff
@@ -0,0 +1,23 @@
+import { selectorFamily } from 'recoil'
+
+export const averageColorSelector = selectorFamily({
+  key: 'averageColor',
+  get: (url: string) => async () => {
+    if (!url || url.startsWith('http://localhost')) {
```
elsehow, line 6: What's going on here? Is this for when it's running on a development server? A comment would help here.

packages/state/recoil/selectors/misc.ts
```diff
@@ -0,0 +1,23 @@
...
+      return undefined
+    }
+
+    try {
+      const response = await fetch(`https://fac.withoutdoing.com/${url}`)
```
elsehow, line 11: This hardcoded URL is sus. Can we move this to a config somewhere? That would let us document better what these endpoints are and do. (Will be helpful when we inevitably do more orchestration around the APIs we depend on).

elsehow (changes requested):

packages/stateless/components/profile/ProfileCardWrapper.tsx
```diff
@@ -28,32 +30,19 @@ export const ProfileCardWrapper = ({
...
+  // Get average color of image URL if in compact mode.
+  const averageImgColorLoadable = useCachedLoadable(
+    !compact || walletProfile.loading
+      ? undefined
+      : averageColorSelector(walletProfile.data.imageUrl)
+  )
+  const averageImgColor =
+    averageImgColorLoadable.state === 'hasValue' &&
+    averageImgColorLoadable.contents
+      ? // If in #RRGGBB format, add ~20% opacity (0x33 = 51, 51/255 = 0.2).
```
elsehow, line 42: Great comment (:

elsehow (approved):

PR #958, opened 2022-12-16
Allow voting on proposals up to completion

0xekez:
Also updates the frontend code IDs to use the [latest release](https://github.com/DA0-DA0/dao-contracts/releases/tag/v2.0.0-beta).

ref: https://github.com/DA0-DA0/dao-contracts/commit/a6e01eacbe3a50ff867d68cf0d2c27aeb4f7f649


packages/stateful/proposal-module-adapter/adapters/DaoProposalSingle/components/ProposalStatusAndInfo.tsx
```diff
@@ -212,6 +212,9 @@ const InnerProposalStatusAndInfo = ({
       : []),
   ]
 
+  const proposalCompleted =
+    proposal.status !== Status.Open && proposal.status !== Status.Passed
```
NoahSaso, line 216: this means that a proposal is only completed (and thus disallows votes cast) when it is rejected, executed(/failed execution), or closed. looking at the contract code, it seems like even passed proposals can be voted on before expiry. am I misunderstanding that, or is there a reason you are preventing passed proposals from being voted on here?

i would think the condition here should be `NOT open AND expired`.

also nit: `proposalCompleted` here is a bit vague as there's additional context relevant here; an explanatory comment might help, or renaming to something like `proposalOpenForVoting`

packages/stateful/proposal-module-adapter/adapters/DaoProposalSingle/hooks/useLoadingWalletVoteInfo.tsx
```diff
@@ -80,18 +80,18 @@ export const useLoadingWalletVoteInfo = ():
     totalVotingPowerWhenProposalCreatedLoadable.contents.power
   )
 
+  const proposalCompleted =
+    proposal.status !== Status.Open && proposal.status !== Status.Passed
```
NoahSaso, line 84: same as other comment

packages/stateful/proposal-module-adapter/adapters/DaoProposalSingle/hooks/useLoadingWalletVoteInfo.tsx
```diff
@@ -80,18 +80,18 @@ export const useLoadingWalletVoteInfo = ():
...
+  const canVote =
+    couldVote && !proposalCompleted && (!walletVote || proposal.allow_revoting)
```
NoahSaso, line 86: love how declarative this is ❤️

packages/utils/constants/codeIdConfigs.ts
```diff
@@ -2,21 +2,21 @@ import { CodeIdConfig } from '@dao-dao/types'
...
+  DaoCore: 3457,
+  DaoPreProposeMultiple: 3460,
+  DaoPreProposeSingle: 3461,
+  DaoProposalMultiple: 3462,
+  DaoProposalSingle: 3463,
+  DaoVotingCw20Staked: 3464,
+  DaoVotingCw4: 3465,
+  DaoVotingCw721Staked: 3466,
+  DaoVotingNativeStaked: 3467,
+  CwTokenSwap: 3456,
```
NoahSaso, line 19: 🥵😫😤🫡 hell yeah

NoahSaso (changes requested):
awesome 🔥 just one thing

packages/stateful/proposal-module-adapter/adapters/DaoProposalSingle/components/ProposalStatusAndInfo.tsx
0xekez: ah you're right. thanks!

NoahSaso:
also I just realized this feature can only activate if the `coreVersion` of the DAO is not V1. (this can be retrieved from the `useDaoInfoContext` hook)

NoahSaso (approved):

PR #959, opened 2022-12-16
Display collection suggestions in mint NFT action, and allow extra metadata

NoahSaso:
This retrieves the list of NFT collections from the DAO when using the mint action, and displays those for which the DAO is the minter. Most of the time, people will be using an existing NFT collection, so we should show them those they've already added to prevent them from having to enter the address every time.

This also allows arbitrary metadata to be added to the JSON file.

Screenshot 2022-12-15 at 9 28 49 PM

Screenshot 2022-12-15 at 9 25 23 PM


JakeHartnell (approved):
Hot.

elsehow (approved):

PR #961, opened 2022-12-17
Move voting from proposal card in sidebar to under the status

### LINKED ISSUE #960: Add 'open sidebar to vote' button under proposal status
To vote, you have to use the profile vote card in the right sidebar. When not in a large desktop window, the sidebar is collapsed and needs to be opened via the profile image circle in the bottom right.

To close and execute proposals, a button appears below the status summary.

When voting is open, we should display a button that opens the sidebar to view the voting interface in the same place. We could also in-line the voting interface, but it might be good to keep that in one place.

NoahSaso:
Resolves #960 

the sidebar voting UI on mobile is a bit out of the way, so let's make it easy for users to vote by moving the vote UI to under the status, which is where the execution and close buttons show as well.

Screenshot 2022-12-17 at 1 20 59 AM
Screenshot 2022-12-17 at 1 21 08 AM

with an action to perform (like execution):
Screenshot 2022-12-17 at 1 23 55 AM
Screenshot 2022-12-17 at 1 24 15 AM


packages/i18n/locales/en/translation.json
```diff
@@ -549,6 +550,9 @@
     "updateProposalSubmissionConfigActionDescription": "Update the proposal submission paramaters for your DAO.",
     "updateVotingConfigActionDescription": "Update the voting parameters for your DAO.",
     "verifyingGovernanceToken": "Verifying CW20 token...",
+    "voteInSidebarLarge": "Vote from the sidebar on the right.",
+    "voteInSidebarResponsiveBody": "You can vote from the sidebar.",
+    "voteInSidebarResponsiveNudge": "Open it by pressing the profile circle in the bottom right or the button below.",
```
elsehow, line 555: Are we still doing this? I thought the voting was now not in the sidebar in responsive mode.

packages/stateful/components/profile/ProfileProposalCard.tsx
```diff
@@ -108,24 +104,18 @@ export const ProfileProposalCard = ({
...
-    <ProfileVotedCard
       {...commonProps}
       vote={
-        // Fallback to none since they can no longer vote.
-        <ProposalWalletVote fallback="none" vote={vote} />
+        <ProposalWalletVote
+          fallback={
+            // If they can vote, fallback to pending to indicate that they still
+            // have to vote.
+            canVote ? 'pending' : 'none'
```
elsehow, line 115: I find `'none'` (vs `None`) a little sketch. I assume it needs to be a string because it's a prop. Maybe something like `'haveNotVoted'` would be more descriptive/less easily confused for a stringified `None`.

packages/stateless/components/NoContent.tsx
```diff
@@ -12,24 +13,43 @@ export interface NoContentProps {
...
+// will navigate to the link. If `onClick` is passed, clicking on this card will
+// execute the function. The `body` text is displayed all the time, whereas
+// `actionNudge` displays below `body` only in the event that an action is
+// present (i.e. one of `href` or `onClick` is set). `buttonLabel` specifies the
+// label of the button that displays below the text, which displays only when an
+// action is present and triggers that action just like clicking on the whole
+// card. The button ensures the user knows an action can be performed, but the
+// whole card remains clickable to improve UX. It is often used when there is no
+// content to prompt the user to perform an action, such as when there are no
+// pinned DAOs.
```
elsehow, line 30: Love this (:

elsehow (approved):
LGTM - a few questions/concerns but nothing in the way of shipping.

PR #962, opened 2022-12-17
Trim average color API response to fix profile card accent opacity

NoahSaso:
The average color API returns text with a newline at the end. This was disrupting the logic used in compact profile cards, such as the one displayed on the proposal page, when adding opacity to the color, since the logic checks to ensure the length of the string is 7 to validate #RRGGBB format before making it #RRGGBBAA.

This trims the API response so the selector just returns the color string, as one would expect, and fixes the opacity logic.

### DIFF
--- a/packages/state/recoil/selectors/misc.ts
+++ b/packages/state/recoil/selectors/misc.ts
@@ -14,7 +14,8 @@ export const averageColorSelector = selectorFamily({
       const response = await fetch(
         FAST_AVERAGE_COLOR_API_TEMPLATE.replace('URL', url)
       )
-      const color = await response.text()
+      // Trim newline at the end.
+      const color = (await response.text()).trim()
       // Validate color format.
       if (!color.startsWith('#')) {
         return undefined


elsehow (approved):
Approved... *and* I'm not crazy about the color API giving a plaintext response. I think APIs should always give structured data. Even if JSON isn't typed, JSON dicts can imply the type. 

For example:

```
{
  'colorHexString': '#ff00ff'
}
```

Is more self documenting than:

```
"#FF00FF"
```

Let's see if we can't make that upstream API JSON... Also, happy to hear counterarguments to this point!

NoahSaso:
> Approved... _and_ I'm not crazy about the color API giving a plaintext response. I think APIs should always give structured data. Even if JSON isn't typed, JSON dicts can imply the type.
> 
> For example:
> 
> ```
> {
>   'colorHexString': '#ff00ff'
> }
> ```
> 
> Is more self documenting than:
> 
> ```
> "#FF00FF"
> ```
> 
> Let's see if we can't make that upstream API JSON... Also, happy to hear counterarguments to this point!

Agreed, typically APIs return structured data. I think @0xekez just threw this together a while ago to solve an immediate need and we didn't spec it out very much, but it's a very easy change. @0xekez would you do that 🙏?

PR #963, opened 2022-12-18
Optionally bypass proposal simulation

### LINKED ISSUE #932: Allow bypassing failed proposal message execution simulation
Right now, you cannot publish a proposal if the actions are not executable by the DAO when you try to publish. In some instances (not most), one may want to intentionally publish a proposal with actions that will become executable in the future, but are not at the time of creation.

We should add a way to bypass the simulation check. I suggest that on attempting to publish a proposal and failing the simulation check, the 'Publish' button becomes 'Publish anyway' with a cooldown of 5 seconds (after which the button should return back to normal). We should also add a comment to the end of the failed error check toast popup to press again to bypass the simulation.

NoahSaso:
Resolves #932 

On v2, we now simulate proposal actions before allowing them to be published to help users catch issues before they open something for a vote.

In some cases, users may want to publish a proposal that will be validly executed later but is not right now. One example is spending money that has not yet been received/claimed but will be at some future time.

This PR adds a 10-second cooldown to let users bypass the simulation when it fails.

image


elsehow (approved):
Love this feature! 

Not sure how strongly I feel about this, but we may want to add a note that says, "Learn more," perhaps with a link to  some docs about how simulation works. Probably not necessary for now; maybe something to consider later on.

PR #964, opened 2022-12-18
Remove established from profile cards

NoahSaso:
The v2 profile card designs contained an "established" date display, which we don't currently use. There isn't a clear use for this date, which indicates that it may not be a very meaningful piece of information. It could mean when you first joined a DAO, when you performed your first transaction, or when you first set your profile name.

This PR removes that code. No need to keep it in until we decide we want it.


elsehow (approved):
great

PR #966, opened 2022-12-19
Save latest proposal form state to localStorage

NoahSaso:
The Create DAO form automatically saves to the browser's local cache to prevent loss of data. Proposals don't have this; instead, they use drafts which let users save incomplete proposals for later.

This PR mimics the create DAO form's behavior of automatically saving the latest state, to prevent accidental data loss. It is scoped by DAO, so there will be a different latest save state per DAO. The save state will get cleared once the proposal is created.

Drafts still exist, though are a slightly different use case. It doesn't hurt to have both.


elsehow (approved):
Beautiful.

PR #967, opened 2022-12-20
Improve IPFS support and fix USDC infinite loading bug

NoahSaso:
This normalizes all image URLs with the IPFS URL conversion utility. It also allows creating DAOs with IPFS images.

Also fixes $USDC price infinite loading bug.

### DIFF
--- a/packages/utils/conversion.ts
+++ b/packages/utils/conversion.ts
@@ -178,9 +178,15 @@ export const transformIpfsUrlToHttpsIfNecessary = (ipfsUrl: string) =>
     ? IPFS_GATEWAY_TEMPLATE.replace('PATH', ipfsUrl.replace('ipfs://', ''))
     : ipfsUrl
 
-// Normalize image URLs by ensuring they are from a valid IPFS provider or not
-// an IPFS URL.
-export const normalizeImageUrl = (url: string) => {
+// Transform image URLs to ensure they can be accessed. They need to be using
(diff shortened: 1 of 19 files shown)


packages/utils/conversion.ts
```diff
@@ -172,8 +172,36 @@ export const durationToSeconds = (blocksPerYear: number, duration: Duration) =>
...
     : duration.time
 
-// Use Stargaze's IPFS gateway.
+// Convert IPFS protocol URL to HTTPS protocol URL using IPFS gateway.
 export const transformIpfsUrlToHttpsIfNecessary = (ipfsUrl: string) =>
   ipfsUrl.startsWith('ipfs://')
     ? IPFS_GATEWAY_TEMPLATE.replace('PATH', ipfsUrl.replace('ipfs://', ''))
     : ipfsUrl
+
+// Normalize image URLs by ensuring they are from a valid IPFS provider or not
```
elsehow, line 181: Is "normalize" really what we mean here? Or is it something like `toValidIPFSUrlWithProvider`?

elsehow (approved):
This is great. Only thing I'm unsure about is whether "normalize" is the best way to name what we're doing.

packages/utils/conversion.ts
NoahSaso: True, renamed and added more comments!

PR #973, opened 2022-12-30
Fix option dark mode styling issue on linux

JakeHartnell:
image

Fixes small issue found in bug club.

To do:
- [x] test to make sure this doesn't mess with dark mode on other OSs


NoahSaso (changes requested):
finally! ❤️

i think we should move this to the `@layer base > :root > .dark` selector (underneath all the colors) so it only activates on dark mode :D

NoahSaso (approved):
feel free to merge once you make that tweak 🙏

PR #978, opened 2023-01-08
USE INDEXER

NoahSaso:
This PR adds the indexer. This should make the site feel as fast as any other site. This also lets us stop relying on SubQuery indexers.

The indexer is not yet documented as it is still being actively developed, and the upcoming release of this v2 UI will be a significant factor in solidifying its features and API. Around the time the v2 release has been deployed, battle tested, and bug-fixed to its final form, the indexer will likely find itself in a similar state and will be documented then.

Here's what that actually means:
- All SubQuery GQL endpoints are replaced by our indexer. This includes the inbox, wallet stats (proposals created & votes cast), DAO search, and the rest.
- All smart contract queries (found in recoil selectors) first attempt to query the indexer for the same data. If no data is returned, if an error is returned, or after a timeout of 3 seconds, we fallback to querying the chain directly. The timeout of 3s is chosen because if the indexer takes longer than that, it is probably experiencing some issue. The chain RPCs are faster than a 3 second indexer query.
- Featured DAOs are bundled together on the indexer and cached so they can be loaded very quickly. Thus the frontend now pulls featured DAOs from the indexer.
- We display the $JUNO price from the Junoswap pools with a 24 hour change percentage. Woooo.
- There is now a Discord Notifier modal on a DAO's home page that allows anyone to receive Discord notifications when new proposals are created.


elsehow (approved):
this is incredible & incredibly thorough. LFG

PR #980, opened 2023-01-09
Fixed choosing between various proposal modules

NoahSaso:
This PR fixes the proposal module dropdown so that it correctly selects the selected proposal module. Right now it selects all of them.

### DIFF
--- a/packages/stateless/pages/CreateProposal.tsx
+++ b/packages/stateless/pages/CreateProposal.tsx
@@ -27,6 +27,8 @@ export const CreateProposal = ({
   const { t } = useTranslation()
   const { RightSidebarContent, PageHeader } = useAppLayoutContext()
 
+  // List of proposal modules available, using the adapter ID to derive a label
+  // to display in the dropdown.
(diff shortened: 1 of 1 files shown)


packages/stateless/pages/CreateProposal.tsx
```diff
@@ -29,15 +29,18 @@ export const CreateProposal = ({
...
-      (
-        daoInfo.proposalModules
-          .map((proposalModule) => matchAdapter(proposalModule.contractName))
-          .filter(Boolean) as ProposalModuleAdapter[]
-      ).map((adapter) => ({
-        label: t(`proposalModuleLabel.${adapter.id}`),
-        value: proposalModule,
-      })),
-    [daoInfo.proposalModules, proposalModule, t]
+      daoInfo.proposalModules
```
elsehow, line 32: This is great, but can we get a comment on here? Something like:

> // Chooses the correct proposal module (i.e., the one that matches the contract name)

Or something like that.

elsehow (approved):

PR #1005, opened 2023-01-11
Clean up inbox

### LINKED ISSUE #993: Show proposals I've already voted on in inbox if they're still open
image

If I've voted on a proposal in a DAO I follow and am a member of, I should see the proposals I've voted on under a heading like "Voted on."

### LINKED ISSUE #994: Hide DAOs with no notifications in inbox
image

When a DAO has no notifications, the DAO can be hidden. For example, in this inbox view, "DAO DAO" could be hidden.

NoahSaso:
Resolves #993 
Resolves #994 

This PR hides DAOs which have no open proposals from showing in the inbox, even if they're followed. It also unhides proposals that the user has voted on; before they were hidden, but that was confusing because they're still open. It's clear from the badge if the user needs to vote, so it's fine to leave them in.

image

### DIFF
--- a/apps/dapp/pages/inbox.tsx
+++ b/apps/dapp/pages/inbox.tsx
@@ -30,26 +30,26 @@ const InnerInbox = () => {
   const pinnedDaoDropdownInfos = useRecoilValue(pinnedDaoDropdownInfosSelector)
 
   const {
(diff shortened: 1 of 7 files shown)


JakeHartnell (approved):
LFG!

elsehow (approved):

PR #1010, opened 2023-01-11
Search profiles in AddressInput component

### LINKED ISSUE #991: More easily enter the addresses of people with profiles
In the wallet input field (see #990), it would be great to more easily search for people I know.

image

For example, I might want to type "noah" in here and have it auto-populate. To check that it's the *right* noah, hovering over the profile might show:

- The wallet's address
- What other DAOs we're both members of (e.g., Development Fund, DAO DAO).

In general, wallet addresses are unusable; we should nudge people away from them, encouraging them to set a profile when possible.

NoahSaso:
Resolves #991 

This PR adds a profile search feature to `AddressInput` components.

I created a new, stateful `AddressInput`, which automatically includes the stateful `ProfileDisplay` component as well as performs the search fetch logic.

https://user-images.githubusercontent.com/6721426/211929093-19f63eb8-5381-4f45-a9de-3edb97a91c12.mov


JakeHartnell (approved):
:exploding_head:

elsehow (approved):
massive

PR #1017, opened 2023-01-12
Improve indexer timeout and fix server side indexer queries

NoahSaso:
This fixes indexer timeout on browsers that don't support the `AbortController`. It also fixes the indexer query during static props which broke when it became a serverless API call.

### DIFF
--- a/packages/state/indexer/query.ts
+++ b/packages/state/indexer/query.ts
@@ -10,28 +10,33 @@ export type QueryIndexerOptions = WithChainId<{
     // Most formulas do not need the time, so make it optional.
     timeUnixMs?: number
   }
+  baseUrl?: string
 }>
 
 export const queryIndexer = async <T = any>(
   type: 'contract' | 'wallet' | 'generic',
   address: string,
   formula: string,
-  { args, block, chainId }: QueryIndexerOptions = {}
+  { args, block, chainId, baseUrl }: QueryIndexerOptions = {}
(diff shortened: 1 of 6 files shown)


packages/state/indexer/query.ts
```diff
@@ -10,28 +10,33 @@ export type QueryIndexerOptions = WithChainId<{
...
-      address,
-      formula,
-      args,
-      block: block ? `${block.height}:${block.timeUnixMs ?? 1}` : undefined,
-    }),
-    headers: {
-      'Content-Type': 'application/json',
-    },
-  })
+  const response = await fetchWithTimeout(
```
elsehow, line 22: Let's get a comment here - what's 3000?

packages/utils/network.ts
```diff
@@ -1,15 +1,37 @@
...
-    clearTimeout(timeoutId)
-  }
-}
+) =>
+  new Promise<Response>(async (resolve, reject) => {
+    const controller = new AbortController()
+
+    let rejected = false
+    const timeoutId = setTimeout(() => {
+      console.log('Aborting...')
```
elsehow, line 10: Aborting what?

elsehow (approved):

packages/utils/network.ts
NoahSaso: This aborts the fetch request below after 3 seconds. Added more info to the log and comments 👍

PR #1020, opened 2023-01-12
Clarify USD estimated value displays

### LINKED ISSUE #1002: DAO cards show misleading treasury values/ambiguous dollar numbers
image

image

In DAO cards, and in the DAO's page, treasury balances are shown by their estimated USDC value. The intention of this design is to give users a sense for the value of a DAO's treausry. However, the way the treasury cards are displayed, users would be forgiven for thinking that DAOs have literal USDC in their treasury. Further, the lack of context around the bank symbol may make some users believe the amount shown is the USDC value of the DAO's governance token!

To clarify, I suggest we instead show "est. USD value" rather than relying on USD per se. In the example above, the card might read:

> $15 est. treasury value.

A tooltip next to this note might read:

> Estimated value of DAO's treasury in US dollars.

NoahSaso:
Resolves #1002 

This PR clarifies when there are $USD estimates, instead of just stating them as $USDC values.

Screenshot 2023-01-12 at 11 53 20 AM
Screenshot 2023-01-12 at 11 53 27 AM

Screenshot 2023-01-12 at 11 53 39 AM
Screenshot 2023-01-12 at 11 53 45 AM

Screenshot 2023-01-12 at 11 53 53 AM
Screenshot 2023-01-12 at 11 53 58 AM


packages/i18n/locales/en/translation.json
```diff
@@ -603,6 +604,7 @@
     "updateInfoActionDescription": "Update your DAO's name, image, and description.",
     "updateProposalSubmissionConfigActionDescription": "Update the proposal submission paramaters for your DAO.",
     "updateVotingConfigActionDescription": "Update the voting parameters for your DAO.",
+    "usdEst": "USD est.",
```
elsehow, line 607: Nit: "est. USD value".

elsehow (approved):
Looks great! Only question: why is there a $ before "USD"? I can't immediately find in the code where that is happening (`symbol` seems none-ish). I think "USD" is more natural than "$USD." Otherwise, looks great!

NoahSaso:
> Looks great! Only question: why is there a $ before "USD"? I can't immediately find in the code where that is happening (`symbol` seems none-ish). I think "USD" is more natural than "$USD." Otherwise, looks great!

Ah, the `format.token` translation sticks a `$` before all symbols. Can change that.

PR #1021, opened 2023-01-12
Improved responsive connect UX, and some other tiny fixes

NoahSaso:
Currently, it is not very clear how to connect your wallet on mobile/small screens. This PR brings the connect wallet button out of the sidebar and replaces the floating empty profile icon with it. It also adds a button that lets the user shrink/hide the connect button if they don't want to connect.

### The old way

Screenshot 2023-01-12 at 1 41 39 PM

### The new way

Screenshot 2023-01-12 at 1 41 50 PM

Clicking the arrow makes it:
Screenshot 2023-01-12 at 1 41 57 PM
image

And clicking the compressed button above brings the connect button back out and starts connecting:
Screenshot 2023-01-12 at 1 42 07 PM

Because the footer is in the right sidebar, I moved it to the navigation sidebar on the left so it's visible when not connected:
Screenshot 2023-01-12 at 1 43 10 PM
Screenshot 2023-01-12 at 1 43 25 PM

Along the way, I also fixed a couple bugs. One required upgrading cosmodal so that the button stops showing loading/connecting if the user closes the wallet selection modal without doing anything, and the other was a strange infinite loading bug in the `DiscordNotifierConfigureModal`.


JakeHartnell (approved):
Think this is a lot better.

PR #1022, opened 2023-01-13
Add governance proposal vote action and cleaned up stargate msg protobuf system

NoahSaso:
This PR adds a governance proposal vote action.

image
Screenshot 2023-01-13 at 3 09 30 AM

In the process, I also:
- cleaned up the protobuf stargate encoder/decoder util functions in `messages.ts` by setting up a `Registry` (from `@cosmjs/*` packages) with all default types loaded and using its `encode` and `decode` functions that match protobuf `type_url` values automatically.
- replaceed the DAO's `coreVersion` in the context object provided to actions with the entire `DaoInfo` type, so that actions have access to the DAO name. Previously, actions that could be used by both a wallet and a DAO could not use the `useDaoInfoContext` hook to retrieve `DaoInfo` because you can't conditionally call hooks, and that hook errors outside of a DAO page (e.g. on a wallet page).
- fixed the light theme text colors of tooltips, toasts, and brightens the pressed button text color a bit.
- added a Wallet page when disconnected with a connect wallet button. I used the wallet page to test validator actions in different contexts.


elsehow (approved):
Amazing.

PR #1023, opened 2023-01-13
Improved estimated USD tooltips.

NoahSaso:
Resolves https://github.com/DA0-DA0/dao-dao-ui/commit/e9ef262d04f0725b0beefcadc6f96e56dcd1607c#r96014249

Changes USD estimate tooltips for DAO treasuries to:

> The USD value of treasuries is estimated by summing the value of all tokens held in the treasury that have a known price on JunoSwap.

...and the following for other USD estimates:

> USD value is estimated using price data from JunoSwap.


elsehow (approved):
perfect!

PR #1030, opened 2023-01-14
Fix authz and tweak action card styles

NoahSaso:
This PR fixes authz action encoders/decoders and touches up the various protobuf encoder/decoder functions to be more general.

I also cleaned up action card styles a bit.

In doing this, we remove the special protobuf decoder overrides because they lead to unexpected behavior, and we don't want to mess with the normal decoding of a given protobuf. Any further decoding should happen after the fact within the specific context, such as in `actions/AuthzAuthorization.tsx` which converts the public key in a create validator message into a readable base64 string. What this means is the CosmWasm message previews will have byte arrays where protobufs are nested inside other messages, which is fine. It's technically more accurate though a bit less helpful.


JakeHartnell (approved):
LGTM!

JakeHartnell (approved):

PR #1040, opened 2023-01-17
Link to DAOs from their displays

NoahSaso:
Reported in Discord #bug-club by sir @0xekez [here](https://discord.com/channels/895922260047720449/941044521909780500/1064718071664738406).

This PR adds a link to open a DAO in a new page from wherever its address is displayed.

image

Crucially, this PR also refactors `ProfileDisplay` with its associated `Profile` type (and related code) to `EntityDisplay` and `Entity`, respectively. The `useProfile` hook and display component served to display wallet profiles only, but was extended to display DAOs names/images. This happened when the token swap action was born, because token swaps necessitated the reconceptualization of DAO addresses and wallet addresses as similar... entities (sorta like how corporations are considered people, except in a much less toxic and problematic way). Both DAOs and wallets are autonomous agents that can perform the same actions and interact with each other. Thus, this profile display really became an entity display, automatically choosing to load a DAO's name/image or a wallet's pfpk profile depending on the input address. Now that wallet profiles are used and referred to all over the place, this system should be refactored to use `entity` verbiage to prevent confusion.


elsehow (approved):
Looks GREAT! Super clear.

JakeHartnell (approved):

packages/stateless/components/inputs/AddressInput.tsx
```diff
@@ -145,18 +145,28 @@ export const AddressInput = <
...
+        'secondary-text group relative flex min-w-0 items-center gap-3 bg-transparent font-mono text-sm transition-all',
+        // If not only displaying entity, add more border.
+        onlyDisplayEntity
+          ? 'p-2'
+          : [
+              'rounded-md py-3 px-4 ring-1 focus-within:ring-2 ',
+              error && !showProfileAutofill
+                ? 'ring-border-interactive-error'
+                : 'ring-border-primary focus:ring-border-interactive-focus',
+            ],
```
0xekez, line 164: should we hide the error outline if `onlyDisplayEntity` is true?

0xekez (approved):
cool!

packages/stateless/components/inputs/AddressInput.tsx
NoahSaso: yep! it's in that array in the else block of the ternary for that reason :)

PR #1057, opened 2023-01-24
Goodbye splash

NoahSaso:
The splash page has served us well... kinda. But not really. Burn it to the ground.

This moves `/home` to `/`.


JakeHartnell (approved):
RIP

PR #1059, opened 2023-01-24
Hide inbox and following when not connected

### LINKED ISSUE #1056: Hide following and inbox in nav when wallet not connected
Now that following DAOs are attached to a wallet login, the inbox and following navigation items are no longer relevant when not signed in. They should probably be hidden or disabled.

NoahSaso:
Resolves #1056 

Since following is now attached to a wallet login, the inbox and following navigation items are always empty when disconnected. Let's just hide them until a wallet is connected.

It looks pretty bare now as a result. I think we'll want to fill in the space with documentation, guides, tutorials, instructions, etc.

Screenshot 2023-01-24 at 1 55 11 PM


JakeHartnell (approved):
LGTM!

PR #1062, opened 2023-01-25
Fix actions refreshing unnecessarily

NoahSaso:
The execute and instantiate actions use native balances to allow specifying funds to be included with the wasm message. Some time after these actions were created, we made the balances update every minute in the background. These actions were using the old way of accessing the balances (i.e. not using a loadable), which caused the components to suspend every minute when the refresh occurs. This is bad UX, and everywhere else we use loadables. This PR updates these actions to use loadables too.

This has bugged me for sooooo long and only on my 9000th time being annoyed by it did I decide to fix it... of course it was age-old tech debt :P


JakeHartnell (approved):
Thanks for this!

PR #1067, opened 2023-01-28
Overhaul tokens & prevent depositing governance tokens

NoahSaso:
This PR overhauls tokens. More specifically, it:
- prevents depositing governance tokens by hiding the Deposit button in the governance token card
- adds manage stake button to governance token card
- adds warning to deposit modal
- standardizes token types
- separates voting module adapter `useGovernanceInfo` hooks from their generalized interface and adds common voting module adapter hook `useCommonGovernanceInfo` to replace `useGovernanceInfo` (`useGovernanceInfo`s were trying to share the same structure across voting module adapters, which made the field names unnecessarily confusing; this emerged because some packages outside the voting module adapters want to access the governance tokens symbol/decimals, regardless of the voting module being used—those packages only use a couple fields, so a simpler generic exported hook can replace the one used internally that was conforming to a bad structure)
- updates `NumberInput` API to remove need for manually specifying `onPlus` and `onMinus`. doing so also standardizes the `onMinus` and `onPlus` functionality.

image
Screenshot 2023-01-28 at 11 43 51 PM


JakeHartnell:
Warning text could read: `Tokens will be sent from your wallet to the DAO.`

JakeHartnell (approved):

PR #1068, opened 2023-01-28
SDA

NoahSaso:
I brought the SDA (from ages past) back to life, in the style of v2. We were previously maintaining the SDA as a separate app, clone-able by other projects but not immediately useful. This SDA is now much more useful, as it provides a DAO DAO view that simply focuses on one DAO, using all the same tabs and UI elements (like navigation) that main DAO DAO uses. This is likely what DAOs will want to give to members to interact with their DAO, since it is more like a dedicated website. Main multi-DAO DAO DAO will likely be used by power users, who want things like the inbox, following list, etc. This single DAO view will be very useful when adoption is wider and more specific, like when businesses run DAOs.

Part of this PR is also the creation of a widget system that allows adding widgets to the home page of a DAO's SDA. Things like Mint to Join, NFT auctions, social features, twitter fees, etc. will likely be very useful in the future. This is not really in use yet, but it is setup to be used the moment we want to :)

Screenshot 2023-01-30 at 3 13 23 PM
Screenshot 2023-01-30 at 3 13 34 PM
Screenshot 2023-01-30 at 3 19 08 PM
Screenshot 2023-01-30 at 3 17 45 PM
Screenshot 2023-01-30 at 3 14 32 PM
Screenshot 2023-01-30 at 3 18 22 PM
Screenshot 2023-01-30 at 3 18 28 PM


JakeHartnell (approved):
LFG! :fire:

JakeHartnell (approved):

JakeHartnell (approved):

PR #1069, opened 2023-01-30
[legacy] Don't error if proposal message fails to decode

NoahSaso:
The WYND legacy multisig created a message that fails to properly decode. This causes the page to crash instead of just showing the raw JSON of the message. This PR fixes that.

https://legacy.daodao.zone/multisig/juno1hrs4jha85xmm56p5z80lq706ckq9zxpfhg3mcx5n7est4ffxykls82eaps/proposals/34


0xekez (approved):
`<ThisIsAReviewBecauseReactIsAnnoying />`

JakeHartnell (approved):

PR #1075, opened 2023-02-03
Added fiat wallet ramp modal and DAO treasury deposit modal.

NoahSaso:
This PR adds fiat modals!

## Wallet ramp

Screenshot 2023-02-03 at 7 39 58 PM
Screenshot 2023-02-03 at 7 40 05 PM

## DAO deposit

Screenshot 2023-02-03 at 5 24 46 PM
Screenshot 2023-02-03 at 5 24 53 PM


apps/dapp/.env.mainnet
```diff
@@ -60,3 +60,6 @@ NEXT_PUBLIC_DISCORD_NOTIFIER_API_BASE=https://discord-notifier.daodao.zone
 
 # Following DAOs
 NEXT_PUBLIC_FOLLOWING_DAOS_API_BASE=https://following.daodao.zone
+
+# Kado API
+NEXT_PUBLIC_KADO_API_KEY=bcee9402-d79f-427b-bafd-dd111e4ce287
```
elsehow, line 65: Just checking: This is okay to publish publicly, right?

packages/i18n/locales/en/translation.json
```diff
@@ -477,6 +478,7 @@
     "daodaoLogo": "DAO DAO Logo",
     "daodaoWithVersion": "DAO DAO v{{version}}",
     "daosLogo": "DAO's Logo",
+    "depositFiatDescription": "Fiat is converted into USDC before it is sent to the DAO's treasury. Balances may take some time to update once the transaction completes.",
```
elsehow, line 481: Not necessarily germane to this PR, but how can users tell the status of their tx? If it completes and the balance isn't updated, is there somewhere they can go to see its status?

elsehow (approved):

apps/dapp/.env.mainnet
NoahSaso: Yep, it's the API key for the iframe, so it's public.

packages/i18n/locales/en/translation.json
NoahSaso: The Kado modal shows the transaction as complete, but it still has to send through the bridge and make it to Juno. So the tx status is on the Kado tx history, and says complete before the DAO treasury shows the token.

JakeHartnell (approved):
:clap:

JakeHartnell (approved):

PR #1078, opened 2023-02-04
Added ability to choose juno NFT for PFP

NoahSaso:
This PR adds Juno NFTs to profile photo picker modal. It also adds the chain image the NFT originates from.

We no longer need to query the Loop API for Juno NFTs since we ask the indexer, so this also removes Apollo/GQL libraries and queries from the codebase.

image


JakeHartnell (approved):
LFG!

PR #1084, opened 2023-02-04
Fixed fiat modal height, scrolling, and address truncation

NoahSaso:
Minor fixes to make scrolling the fiat modal better on mobile.

Also unrender the fiat iframe when it is not visible because iframe hijacks touches on mobile Safari even when hidden :////////////
huge bruh moment


onewhiskeypls (approved):
**utACK**: lgtm

JakeHartnell (approved):

PR #1111, opened 2023-02-14
WYND Swap Action, prices from WYND, and token refactor

### LINKED ISSUE #1097: Search DAOs in address inputs
As of #990, address inputs let you search for wallets using [pfpk](https://github.com/DA0-DA0/pfpk).

This should also include DAO search since address inputs usually support both! Address inputs have a `type` (`wallet` or `contract`) which determine the icon to show. I think that should be generalized to specify what kinds of addresses to support, and then intelligently allow searching for those only. For example: some inputs (like registering a SubDAO) are only relevant to enter DAOs into, not wallets. Thus, it should support one or both of wallets and DAOs.

### LINKED ISSUE #1105: Missing Option for Funding Smart Contract Execution with CW20
**Suggestion for improvement:**
Missing Option for Funding Smart Contract Execution with CW20

**Steps to Reproduce:**
Attempt to fund the execution of a smart contract
Observe the available options for funding the execution

**Expected Result:**
There should be an option to fund the execution of a smart contract with a CW20 token.

**Actual Result:**
There is no option to fund the execution of a smart contract with a CW20 token.
image

NoahSaso:
This PR adds an action that lets you perform a token swap with WYND's pools.

It also:
- replaces price calculations from Junoswap with WYND's price API
- makes the actions renderer update less often, using a deep compare to update the actions form instead of comparing references which change more often
- makes the proposal page responsive at a slightly larger width so that actions aren't as shrunk
- memoizes cached loadables to prevent unnecessary re-renders and occasionally render loops
- refactors the token types and selectors, consolidating them into a common GenericToken type (along with a genericTokenSelector) that supports both native and cw20 tokens
- allows executing smart contracts with cw20s instead of just native tokens (resolves #1105)
- allows searching DAOs in AddressInput fields (resolves #1097)

image
Screenshot 2023-02-20 at 6 17 24 PM
Screenshot 2023-02-14 at 8 12 57 PM

### When minimum output is none

Screenshot 2023-02-20 at 6 20 48 PM

### Once created

Screenshot 2023-02-20 at 6 19 18 PM


packages/i18n/locales/en/translation.json
```diff
@@ -559,10 +561,13 @@
...
+    "maxSlippageDescription": "As the transaction is executing, the swap price will fluctuate. If the price slips beyond this maximum slippage, the swap will be invalidated.",
     "members": "members",
     "membershipDaoNotMemberInfo_dao": "You are not a member of {{daoName}}. To become a member, existing members must vote to add you.",
     "membershipDaoNotMemberInfo_proposal": "You were not a member of {{daoName}} when this proposal was created, so you cannot vote on it.",
     "migrateSmartContractActionDescription": "Migrate a CosmWasm contract to a new code ID.",
+    "minimumOutputRequiredDescription_dao": "Before the proposal is passed and executed, the swap price will fluctuate. If the price drops and no longer satisfies this minimum output required, the swap will be invalidated.",
+    "minimumOutputRequiredDescription_wallet": "The exact swap price will fluctuate during the transaction, but the minimum output amount is guaranteed. If the price drops and no longer satisfies this minimum output required, the swap will be invalidated.",
```
elsehow, line 570: "invalidated" is a little odd. Perhaps:

>  If the price drops and no longer satisfies this minimum output required, the swap will not occur.

packages/state/recoil/selectors/wynd.ts
```diff
@@ -0,0 +1,104 @@
...
+    } catch {
+      return undefined
+    }
+  },
+})
+
+// Price of 1 macro token of the specified denom or address.
+export const wyndUsdPriceSelector = selectorFamily<
+  AmountWithTimestampAndDenom | undefined,
+  // denomOrAddress
```
elsehow, line 44: what's this `denomOrAddress`? That's not immediately intuitive to me.

elsehow (approved):
LFG!

packages/state/recoil/selectors/wynd.ts
NoahSaso: denom refers to a native denom, such as `ujuno`, and address refers to a CW20 contract address. This terminology is used in many many places, and it also how WYND serves its prices (though it's just a string labeled `asset` in their API). `denomOrAddress` is commonly used throughout the UI so I think it's consistent 🤷

PR #1120, opened 2023-02-17
Fix early button loading finish, and update revoting button text

NoahSaso:
When voting on, executing, and closing a proposal, there is a small delay between when the UI detects the action occurred/shows an alert, and when the data is refreshed to display. This PR makes it so that the buttons keep loading until the data is refreshed, instead of just waiting until the TX completes.

It also prevents the user from re-casting a vote if they already voted, since that would lead to infinite loading per the system above. In doing so, I changed the button to `Change your vote` when revoting, so it's clear what's going on, and it's clear why the button is disabled when selecting your current vote.

### DIFF
--- a/packages/i18n/locales/en/translation.json
+++ b/packages/i18n/locales/en/translation.json
@@ -15,6 +15,7 @@
     "cancel": "Cancel",
     "castYourVote": "Cast your vote",
     "change": "Change",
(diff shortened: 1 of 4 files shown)


elsehow (approved):
ugh excellent

PR #1125, opened 2023-02-17
Clean up AppContext refs

NoahSaso:
A while back I discovered React portals that lets you transport components to other places in the DOM. This has been used so that pages can control what profile card appears in the sidebar as well as what content appears in the page header, since a layout component manages navigation, header, sidebar, and page content at the same level.

When I did this, I built it an unnecessarily complex way, and for some reason that only started causing problems right now. I went back, questioned my past sanity, and fixed the code so it doesn't do weird insane memoized function component creation shit. Look at this PR, the difference is funny.


onewhiskeypls (approved):
**utACK**: lgtm

PR #1126, opened 2023-02-17
Fix page header invisible on load sometimes

NoahSaso:
Since refs do not trigger re-renders, the page currently doesn't load in with the header showing, which looks weird. This PR triggers an immediate re-render when the page header ref and right sidebar ref are set to fix this. In the previous change I made (#1125), I did not realize that the state variables served that purpose of causing a re-render. Now it all seems to work reliably....


baoskee (approved):

onewhiskeypls (approved):
utACK: lgtm

PR #1127, opened 2023-02-18
SDA Header Fix

NoahSaso:
For some reason, the SDA header is still hiding on load sometimes. Right now it's only happening on mainnet SDA, not on the testnet, and not on the dApp. This is weird.

I suspect it has something to do with the DaoPageWrapper that wraps the whole layout suspending the layout, which makes it take an extra render cycle to load the PageHeader in and set the ref. This could be explained by the testnet loading so fast that the flicker is imperceptible, whereas the dApp takes longer to load data. On the dApp, the layout is not suspended because the DaoPageWrapper depends on the page, so it doesn't have this problem.

Suspending the SDA page inside the layout seem to fix this problem...

### DIFF
--- a/packages/stateful/components/SdaLayout.tsx
+++ b/packages/stateful/components/SdaLayout.tsx
@@ -17,6 +17,7 @@ import {
   DaoCreatedModal,
   InstallKeplrModal,
   NoKeplrAccountModal,
+  PageLoader,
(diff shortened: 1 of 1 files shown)


onewhiskeypls (approved):
utACK: lgtm

elsehow (approved):

PR #1129, opened 2023-02-18
Generalize WebSockets, update inbox live, and improve performance

NoahSaso:
This PR makes the WebSocket a global instance stored in recoil state instead of a hook in the `AppContextProvider`. It also adds more web socket hooks so you can subscribe to an event from many different channels (such as various DAOs). This is used to subscribe to all following DAOs in the inbox so the inbox updates live with every proposal update.

The `AppContextProvider` no longer provides the web socket through its context, but it is now responsible for unsubscribing from channels that no longer have listeners. We can't unsubscribe from channels when the listener hook unmounts, since a channel may be subscribed to from any place in the app, and the hooks don't know about each other. To solve this, each hook adds to a global counter of channel subscriptions, and a central component waits for the subscriptions to reach 0 before unsubscribing.

This also fixes some performance issues.


JakeHartnell (approved):

elsehow (approved):
great, and fantastic docs

PR #1137, opened 2023-02-22
Speed up profile loading

NoahSaso:
Thanks to pupmos pointing out to me how bech32 hashes relate to public keys, I updated pfpk (https://github.com/DA0-DA0/pfpk/pull/8) to allow querying by bech32 address or hash. This means we no longer have to query the chain for an address' public key and can go straight to pfpk!!!!!

I took this opportunity to completely refactor profiles, making them load filler data and update as profile data is ready.


JakeHartnell (approved):

JakeHartnell (approved):

elsehow (approved):
oy fantastic.

PR #1138, opened 2023-02-23
Fix incorrect balance display

NoahSaso:
Fix minor bug showing wrong max balance in WYND action. Also revalidate when simulation updates amount.


onewhiskeypls (approved):
utACK: lgtm

PR #1159, opened 2023-03-06
Detect Ledger format error

NoahSaso:
This detects the common error with symbols in the proposal title/description when using a Ledger and shows a more descriptive error message.

image


elsehow (approved):
Fantastic

PR #1161, opened 2023-03-06
Upgrade to Next v13 and React v18.2.0

NoahSaso:
This upgrades to the latest Next.js and React versions! It went much smoother than expected...

(I also fix some profile image URLs not being converted from IPFS.)


elsehow (approved):

NoahSaso:
The Next.js v13 dev server does not work with circular dependencies apparently (even though Next.js v13 builds production fine...). Next.js v12 dev server does work and has been working, and we have many circular dependencies... removing all of them would require rearchitecting the frontend, which is not worth it at the moment.

Circular dependencies are safe as long as imports are referenced in functions and not in the global scope. As most of the code is React components and hooks, this is fine. Though long term this is not ideal. For now, I'm closing this and will upgrade React to v18.2.0 by itself.

PR #1175, opened 2023-03-14
Support NFTs with audio and video

NoahSaso:
This allows creating and viewing NFTs with audio and video content.

During creation:
Screenshot 2023-03-14 at 7 31 28 PM

In proposal view:
Screenshot 2023-03-14 at 3 10 03 PM

On treasury page:
Screenshot 2023-03-14 at 3 10 58 PM
Screenshot 2023-03-14 at 3 11 23 PM

I tried to model an audio player similar to the native video component.


packages/i18n/locales/en/translation.json
```diff
@@ -317,6 +318,7 @@
     "createValidatorMessage": "Create validator message",
     "createValidatorMessageTooltip": "Create a validator controlled by your DAO. Note: you will need to run a validator node (on a server for example). Be sure to use that node's public key (junod tendermint show-validator).",
     "currentValue": "Current value",
+    "customJson": "Custom JSON",
```
elsehow, line 321: What's this for? I think we need at least a tooltip here...

packages/stateful/actions/actions/nft/makeMintNftAction.tsx
```diff
@@ -149,6 +149,8 @@ export const makeMintNftAction: ActionMaker<MintNftData> = ({ t, address }) => {
       name: '',
       description: '',
       includeExtra: false,
+      audio: undefined,
```
elsehow, line 152: Why `undefined` instead of `null`?

elsehow (approved):
This is incredible. I have a few minor questions. Otherwise, LGTM.

packages/stateful/actions/actions/nft/makeMintNftAction.tsx
NoahSaso: `undefined` gets removed from objects when they're stringified, and `null` does not. so i generally choose undefined when filling in default empty values :) actions are duplicated and their data fields are stringified into the URL params, so that makes it a bit shorter, for example

packages/i18n/locales/en/translation.json
NoahSaso: this is just if you want to add any extra properties to the NFT token_uri `metadata.json`. `audio` and `video` end up there as well, along with name, description, and image. I will add a tooltip 👍

elsehow: I wonder if that could be abused... Like, could an attacker add props to do frontend code injection? Also, why bother with this at all? What's a usecase?

NoahSaso: Hmm, if someone *can* do frontend code injection with the metadata of an NFT, they can do it without this field being present (likely they would do it offline on their own wallet and then send it to a DAO). That would require the native json parser to have a vulnerability that allowed for code injection, which we have to use for decoding messages and already have no control over.

NFT traits, for example, are stored in the metadata. So if someone is trying to replicate an NFT with characteristics and what not, they would do it here. I imagine if someone runs a DAO on DAO DAO for some other platform that supports more advanced NFTs, this may be useful. Though on DAO DAO there is no usecase since we only support those additional `audio` and `video` properties for now.

NoahSaso: I'd be happy to remove it for now if you think that's a good idea and just leave the audio and video fields under the extra metadata switch.

JakeHartnell (approved):
HOT

packages/i18n/locales/en/translation.json
elsehow: > Hmm, if someone can do frontend code injection with the metadata of an NFT, they can do it without this field being present ... That would require the native json parser to have a vulnerability that allowed for code injection

Very true.

I'd keep it more restrictive for now, if only because (1) it's slightly confusing and (2) no one's asking us for more fields... If you think it's plausible that someone would, or know some specific usecases off the top of your head, leave it in. Otherwise, let's remove it until someone asks.

sascha1337: if the ui would be forked and some amateur dev would straight append some data to the DOM, unescaped unprotected ? 99% sure this indeed would be some extreme dangerous XSS vulnerabity, depending on the modified fork wrecking people into ruin, thanks for removing this

NoahSaso: > > Hmm, if someone can do frontend code injection with the metadata of an NFT, they can do it without this field being present ... That would require the native json parser to have a vulnerability that allowed for code injection
> 
> Very true.
> 
> I'd keep it more restrictive for now, if only because (1) it's slightly confusing and (2) no one's asking us for more fields... If you think it's plausible that someone would, or know some specific usecases off the top of your head, leave it in. Otherwise, let's remove it until someone asks.

no one has asked us for this, best reason to not add unnecessary complexity :)

PR #1179, opened 2023-03-16
Fix duplicate tokens being counted in token card

NoahSaso:
Reported in #bug-club in the Discord: https://discord.com/channels/895922260047720449/941044521909780500/1085754004484673537

Right now, DAOs that use the same CW20 staking contract are double counting tokens in the token card:
image

This fixes that bug.


elsehow (approved):
great, love a good reduce

JakeHartnell (approved):

PR #1180, opened 2023-03-18
More authz

### LINKED ISSUE #1176: Support more Authz Features
Authz now supports smart contracts! The actions UI for Authz should support them as well! This allows DAOs to authorize certain interactions with smart contracts (including calling specific messages or allowing for migration)! 

In addition, `authz` has supported spend operations for a while now. Would be good to support those too.

JakeHartnell:
Closes #1176


packages/utils/messages.ts
```diff
@@ -256,6 +286,97 @@ export const decodeStargateMessage = ({
...
+        },
+      })
+    }
+  } else if ('stargate' in msg) {
+    encoded = {
+      typeUrl: msg.stargate.type_url,
+      value: fromBase64(msg.stargate.value),
+    }
+  } else {
+    console.error('Msg type not supported')
```
JakeHartnell, line 375: Not sure what we want to do here if it's an unsupported message type. Think I covered all the options, but maybe we should throw an error?

packages/utils/messages.ts
```diff
@@ -179,8 +196,28 @@ export const typesRegistry = new Registry([
   // Custom types not in @cosmjs/stargate default registry.
   ...([
     ['/cosmos.slashing.v1beta1.MsgUnjail', MsgUnjail],
+    ['/cosmos.authz.v1beta1.MsgGrant', MsgGrant],
+    ['/cosmos.authz.v1beta1.MsgExec', MsgExec],
     ['/cosmos.authz.v1beta1.GenericAuthorization', GenericAuthorization],
+    ['/cosmos.authz.v1beta1.Grant', Grant],
```
NoahSaso, line 202: Screenshot 2023-03-23 at 11 43 04 PM

I believe the 3 `authz` messages you added are included by default in the registry types from line 193. only `GenericAuthorization` needs to be manually added here from `authz`

packages/utils/messages.ts
```diff
@@ -179,8 +196,28 @@ export const typesRegistry = new Registry([
   // Custom types not in @cosmjs/stargate default registry.
   ...([
     ['/cosmos.slashing.v1beta1.MsgUnjail', MsgUnjail],
+    ['/cosmos.authz.v1beta1.MsgGrant', MsgGrant],
+    ['/cosmos.authz.v1beta1.MsgExec', MsgExec],
```
NoahSaso, line 200: Screenshot 2023-03-23 at 11 43 04 PM

I believe these 2 `authz` messages are included by default in the registry types from line 193

JakeHartnell:
Not sure why I'm getting that build error.

packages/utils/messages.ts
NoahSaso: this fn does the same thing as `cwMsgToEncodeObject`, which we use to simulate
proposal messages :)

NoahSaso (approved):

Issue #1151, opened 2023-02-28
Add member CSV download to members tab


NoahSaso:
Proposals have a link at the bottom right that downloads the votes in a CSV format:
![Screenshot 2023-02-27 at 7 52 12 PM](https://user-images.githubusercontent.com/6721426/221749224-0546b4a7-6ba2-49a2-a5b7-cf36c58b455b.png)

The new membership tab should have the same thing for members and their voting powers.

NoahSaso:
Closed by #1184

PR #1184, opened 2023-03-22
Add member CSV download to members tab

sagarsaija:
Issue: 
https://github.com/DA0-DA0/dao-dao-ui/issues/1151

This adds CSV download functionality for members.

Screenshot:
Screenshot 2023-03-21 at 5 55 04 PM

Demo:
memberscsv2


NoahSaso (changes requested):
sweet! thank you.

i'm thinking it kinda disrupts the flow being right under the title, could you move it to below the pagination, between lines 359 and 360? this will also ensure that it doesn't show up when there are no members or when the members fail to load :D

PR #1188, opened 2023-03-24
v2.1 fixes and upgrades

NoahSaso:
This upgrades @noahsaso/cosmodal, @cosmjs/\*, and @keplr-wallet/\* to their latest versions.

It also fixes the v2 upgrade action that should not be showing on v2 DAOs but is, and adjusts MC title spacing a bit.


elsehow (approved):
sick

PR #1191, opened 2023-03-25
Add balances to membership tab and CSV download

### LINKED ISSUE #1142: Add number of tokens in members tab
Right now, the members tab shows percent voting power for each member. For tokens and NFTs, we should also show how many tokens / NFTs are owned and staked, in addition to voting power.

NoahSaso:
Resolves #1142 

This adds staked balances and voting weights to the members tab and CSV download.

Screenshot 2023-03-25 at 2 06 32 PM

Screenshot 2023-03-25 at 2 06 13 PM


elsehow (approved):
fantastique

Issue #1199, opened 2023-03-28
Issue: Inability to Add Both Payroll and Retroactive Payments as Payment Options for DAO


ShayneRarma:
Steps to Reproduce:

Add payroll and retroactive payments as payment options for a DAO
Notice that only one widget is available for these payment options
Attempt to add both payment options as separate widgets
Expected Result:
The ability to add both payroll and retroactive payments as separate payment options for a DAO using separate widgets.

Actual Result:
Only one widget is available for adding payroll and retroactive payments as payment options for a DAO. This limits the ability to use both payment options in the same DAO.

Can verify in Rarma DAO - https://daodao.zone/dao/juno1nvfw3xesfz5s8k8l85kuwra5dcryzheu4nq2m2yfe79l9tc0zx9sju5acm

NoahSaso:
Closed by #1207

PR #1207, opened 2023-04-03
Merge payroll adapter into widgets

NoahSaso:
This PR consolidates some architecture and solves the following issues:
- only one payroll system can be used at a time (vesting or retroactive) (#1199)
- enabling/updating widgets requires manual JSON entry which is prone to errors
- all widgets show on the home page
- there is no way to create arbitrary custom tabs for a DAO
- we are currently on a trajectory to increase the number of distinct 'systems' (e.g. payroll) that exist when essentially they're all just variations of custom UI features and can share an API

The solution:
- turn both payroll systems into widgets
- add a location field to widgets that specifies whether they show up on the homepage or in their own tab (eventually in the header as well, for the social widget)
- add an editor component to widgets that slots into the manage widgets action (since some widgets will require more complex setup, such as creating the vesting manager contract with a wallet first)
- allow enabled widgets to add actions to the proposal interface

Essentially, many widgets can be used simultaneously, so the payroll systems should be widgets, and widgets should become an abstraction around any custom UI. 

Right now, to enable payroll, there is a nice, clear `Manage Payroll` action. That is going away now that these are widgets, but we don't want people to have to use the widgets action to enable payroll, as it may be confusing or make it seem less of a Thing. To preserve the UX of choosing your payroll system, I added two new actions that essentially just pass the Manage Widgets action logic through but are more to the point:
image


packages/i18n/locales/en/translation.json
```diff
@@ -1204,10 +1210,14 @@
   },
   "widgetDescription": {
     "mint_nft": "Display an NFT collection with a button to mint new NFTs.",
+    "retroactive": "After each contribution cycle, contributors are assigned points based on their contributions. Payment is split based on the points they receive.",
```
elsehow, line 1213: > Reward contributors for work they've already done. Vote on how to distribute payments in regular cycles.

packages/i18n/locales/en/translation.json
```diff
@@ -1204,10 +1210,14 @@
...
+    "vesting": "A vesting contract distributes funds to contributors over time at a constant rate. The DAO can vote to cancel vesting at any time.",
```
elsehow, line 1214: > Vest funds to contributors at a constant rate, similar to a salary.

elsehow (approved):

PR #1208, opened 2023-04-03
Use chain registry for IBC assets

NoahSaso:
This replaces the local JSON of IBC assets we have to maintain with Cosmology's [`chain-registry`](https://github.com/cosmology-tech/chain-registry/tree/main/packages/juno)!


elsehow (approved):
lovely

PR #1213, opened 2023-04-05
Fix DAO Admin Exec not properly loading DAO context

NoahSaso:
This PR fixes the DAO info selector not properly loading items from the DAO. These items are currently only loaded to detect whether or not to show certain widget actions (to check if they're already enabled), and this specific selector is only used by the DAO Admin Exec action, so it's not a big deal, but it is a bug.


elsehow (approved):
hell yeah

PR #1215, opened 2023-04-05
Fix action amount docs.

NoahSaso:
This removes inaccurate amount/denom explanations from action docs. I previously added clarifications that `amount` was in the `denom` base unit with no decimals, but every action actually converts the data, so this is not true.

### DIFF
--- a/packages/stateful/actions/core/smart_contracting/Execute/README.md
+++ b/packages/stateful/actions/core/smart_contracting/Execute/README.md
@@ -30,8 +30,3 @@ guide](https://github.com/DA0-DA0/dao-dao-ui/wiki/Bulk-importing-actions).
 
 If `cw20` is true, then `funds` can only contain 1 entry, and `denom` should be
 the address of a CW20 token.
-
-`amount` and `denom` are in the base unit with no decimals. For example, if you
-want to spend 1 $JUNO, since JUNO has 6 decimals, then `amount` should be
-`1000000` and `denom` should be `ujuno`. If this is confusing, please ask for
-help!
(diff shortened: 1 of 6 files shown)


elsehow (approved):
This is good, but kind of annoying that the docs aren't DRY. Could we write something in the wiki and link to it throughout the docs? Or is there some reason why we don't want to do that?

JakeHartnell (approved):

elsehow:
I hear that. I do like that it's self-explanatory.

The main reason for DRYer docs in my mind is maintainability. Having this information scattered around works great when you're the one managing the codebase, and you know everything about it. Someday, when more people are working on it, it's conceivable that someone may change the behavior in one part of the code and not know these docs (and thus the patterns) appear elsewhere. 

Also, someday, people will submit their own actions. Is it up to them to decide how `amount` and `denom` work, or should they play along with the way it works in the rest of the app? Probably the former - which makes me feel the pattern should be "enforced" in some single place---perhaps in an Actions wiki?

I don't feel super strongly about this one way or the other. Ultimately, it's up to whatever's easiest to maintain. I do appreciate that this information is close at hand - but, whenever I see stuff substantially duplicated, I always feel there's a risk of something going wrong.

NoahSaso:
> I hear that. I do like that it's self-explanatory.
> 
> The main reason for DRYer docs in my mind is maintainability. Having this information scattered around works great when you're the one managing the codebase, and you know everything about it. Someday, when more people are working on it, it's conceivable that someone may change the behavior in one part of the code and not know these docs (and thus the patterns) appear elsewhere.
> 
> Also, someday, people will submit their own actions. Is it up to them to decide how `amount` and `denom` work, or should they play along with the way it works in the rest of the app? Probably the former - which makes me feel the pattern should be "enforced" in some single place---perhaps in an Actions wiki?
> 
> I don't feel super strongly about this one way or the other. Ultimately, it's up to whatever's easiest to maintain. I do appreciate that this information is close at hand - but, whenever I see stuff substantially duplicated, I always feel there's a risk of something going wrong.

Ah yeah, makes total sense. I was confused because this PR is making things DRYer by deleting information, but technically the amount and denom still follow a pattern, just the more intuitive one. That should probably be codified somewhere...

PR #1219, opened 2023-04-05
Paginate actions if there are many

NoahSaso:
Now that we support bulk actions, it's possible proposals will contain many actions, often of the same type, and make the proposal page very long / slow down having to load a bunch of state into memory.

This PR makes the action card paginate once it has more than 20 actions, and display the total number of actions it contains if more than 1. It also displays a warning calling out that there are multiple pages and to reminding to check each one.

Screenshot 2023-04-05 at 4 15 20 PM
image

This also requires that one views all action pages before voting, for safety:
image

I also added an explanation for why you can still vote on a proposal once the vote is determined:
image


elsehow (approved):

JakeHartnell:
Hmmmmmmmm, I'm not sure about this. Seems like a great way to sneak in something malicious. :thinking: 

It does _look_ nicer.

NoahSaso:
> Hmmmmmmmm, I'm not sure about this. Seems like a great way to sneak in something malicious. 🤔
> 
> It does _look_ nicer.

that's a very good point 🤔 maybe instead the actions part can have a max height scroll internally? instead of extending the page length. wdyt?

elsehow (approved):
Great warning.

PR #1221, opened 2023-04-05
Add self-rating to retroactive compensation system

NoahSaso:
This adds fields to the retroactive compensation submission form where a contributor can state what they would like to be compensated for their submission. It then displays this feedback to the raters while they are rating submissions.

On submission:
image

While rating:
Screenshot 2023-04-05 at 3 51 17 PM

On nomination:
Screenshot 2023-04-05 at 2 37 56 PM


elsehow (approved):
This looks great. My only note: can the "how they would like to be compensated" appear in the far-right column to the raters, and perhaps with a different (secondary) bg on it? It would make it easier to compare side-by-side, I think.

Otherwise, this is fantastic. Can't wait to use this next cycle.

PR #1226, opened 2023-04-06
Preserve action order within groups

NoahSaso:
PUPMOS pointed out that the action groups disrupt the order of actions, but the order actions are executed in *is very important*.

This PR fixes that by only grouping adjacent actions of the same type (instead of pulling actions into an earlier group), preserving action order, while still saving space by grouping actions.

## On creation
image

## On viewing
Screenshot 2023-04-06 at 2 45 09 AM


elsehow (approved):
great catch by pupmos

PR #1372, opened 2023-09-07
Update to latest cosmos-kit versions and add more wallets

NoahSaso:
This adds support for:
- Metamask Snaps wallet
- Fin wallet
- OKX wallet


JakeHartnell (approved):
:fire:

JakeHartnell:
Hmmmm... getting:
```
The method "wallet_requestSnaps" does not exist / is not available.
```

Maybe because snaps are only in development Metamask until later this month?

JakeHartnell:
> Maybe because snaps are only in development Metamask until later this month?

If this is the case, it probably should be lower down the list?

NoahSaso:
> Hmmmm... getting:
> 
> ```
> The method "wallet_requestSnaps" does not exist / is not available.
> ```
> 
> Maybe because snaps are only in development Metamask until later this month?

Yeah it doesn't work yet.

NoahSaso:
> > Maybe because snaps are only in development Metamask until later this month?
> 
> If this is the case, it probably should be lower down the list?

I don't think we should merge this until it's ready.

PR #1394, opened 2023-10-03
Support active threshold in creation and display on DAO page

### LINKED ISSUE #801: Add support for Active Threshold option in token based DAOs
Should default to something sensible like 10%.

Otherwise, there is a race to stake scenario.

NoahSaso:
Resolves #801 

On creation:
Screenshot 2023-10-03 at 14 57 24

When inactive:
Screenshot 2023-10-03 at 14 56 13


JakeHartnell (approved):
🥳

PR #1418, opened 2023-10-14
New Widget: Wrappr 

hard-nett:
Hey! Opening up this PR to track the possible addition of [Wrappr.wtf](https://wrappr.wtf). 

### Whats added
- Basic UI Component layout
- Basic WrapprData Types

### Whats still needed
- middleware to call solidity contract 
- any additional, helpful ui components


JakeHartnell:
Wow! Really cool! @NoahSaso have a look. :eyes:

JakeHartnell:
Vercel builds are failing, if we could fix those it would make it easier to test and try out.

discoverdefiteam:
@JakeHartnell hey! still a decent amount of work needs to be done in order for us to test a basic mvp of this widget(mint a wrappr to a DAO treasury), this dao on testnet is one I have been using as we are building this out:
`juno1v9j7fghulza05jv5sg57q3u6rzlqps2hgxfg55mn5a237j8erg0sqtv8f8`.

took a minute to write down what it seems like we still will need in order to get this to work:

- [ ] handle form data for creating new wrappr; based on entity & jurisdiction
- [ ] check for name collision *(ref: https://github.com/kalidao/wrappr-ui/blob/main/src/create/checkName.ts)*
- [ ] handle displaying PDF of wrappr templates *(will need kaliDAO & LexDAO to update pdf verbiage to reflect CosmWasm contracts, but for now can use existing ones as placeholders)*
- [ ] enforce global mint fee to dao-dao treasury

 unsure if updating llc agreements is the move, might be better to have a dao propose to discard a wrappr agreement, and then mint a new one. May also be handy to enforce only 1 wrappr token per dao treasury, to avoid confusion.

hard-nett:
closing this one for now, going to take a different approach than what we have here :)

PR #1442, opened 2023-11-02
Amino encoding fixes

### LINKED ISSUE #1269: Upload profile image (using instantiate2) doesn't work for ledger/walletconnect
The default Amino types register doesn't have MsgInstantiateContract2, so the upload image does not work.

image

### LINKED ISSUE #1400: Authz Grant simulation error
When trying to use the Authz Grant action, the simulation fails with the following error:
`Query failed with (6): rpc error: code = Unknown desc = unable to resolve type URL : tx parse error [cosmos/cosmos-sdk@v0.47.4/x/auth/tx/decoder.go:42] With gas wanted: '0' and gas used: '1144' : unknown request`

However, when publishing the proposal anyway and executing it, execution succeeds. This indicates that there is an issue in the simulation or in the encoding for the simulation (likely the encoding because protobufs are so tricky).

### LINKED ISSUE #1427: Actions on Chain Governance page are not full width
Doesn't look as good as on normal DAOs.

image

Pardon my zoomed out screen, but it would be nice if they were the full width of the column.

NoahSaso:
Resolves #1269
Resolves #1400
Resolves #1427
Related to https://github.com/cosmology-tech/telescope/issues/506 and https://github.com/cosmology-tech/telescope/issues/508

This fixes all the Amino signature verification failures users have been encountering, including authz, instantiate2, and v1 gov prop submission. Now ledgers can use profile pictures and submit gov props.

This also fixes a tiny bug with the Spend action when trying to use a chain with an invalid IBC destination. And it adjusts the action card layout to be a bit cleaner.


JakeHartnell (approved):
:partying_face:

Issue #1445, opened 2023-11-06
Remove price from claimed balance line of vesting payment


NoahSaso:
Once funds are claimed, the price displayed is misleading because it always uses the current price. I think we can just hide the price for the claimed balance.

rishav-karanjit:
Hi @NoahSaso, I was wondering if I can get few more details on this issue?

NoahSaso:
> Hi @NoahSaso, I was wondering if I can get few more details on this issue?

Yep! This "Claimed balance" line in the vesting payment card is misleading:
![image](https://github.com/DA0-DA0/dao-dao-ui/assets/6721426/26fecc62-9a0b-4cce-ba4a-a8f28504d198)

rishav-karanjit:
Hello @NoahSaso, after heavy search I found this code [here](https://github.com/DA0-DA0/dao-dao-ui/blob/development/packages/stateful/widgets/widgets/VestingPayments/components/VestingPaymentCard/VestingPaymentCard.tsx#L486-L523). Is this the code related to the issue? If not, could you point me to the file related to this issue?

NoahSaso:
Yes that's it! I think all you need to do is delete the price display piece :)

rishav-karanjit:
I have create a PR, @NoahSaso.

rishav-karanjit:
@NoahSaso I updated the PR by just removed the USD price

PR #1450, opened 2023-11-08
Remove the USD price from claimed balance on the vesting card

rishav-karanjit:
For issue #1445

- PR to remove the USD price in claimed balance

### DIFF
--- a/packages/stateful/widgets/widgets/VestingPayments/components/VestingPaymentCard/VestingPaymentCard.tsx
+++ b/packages/stateful/widgets/widgets/VestingPayments/components/VestingPaymentCard/VestingPaymentCard.tsx
@@ -483,44 +483,6 @@ export const VestingPaymentCard = ({
             </div>
           )}
 
-          <div className="flex flex-row items-start justify-between gap-8">
-            <p className="link-text">{t('title.claimedBalance')}</p>
-
-            {/* leading-5 to match link-text's line-height. */}
-            <div className="caption-text flex flex-col items-end gap-1 text-right font-mono">
-              {/* leading-5 to match link-text's line-height. */}
-              <TokenAmountDisplay
-                amount={claimedAmount}
-                className="leading-5 text-text-body"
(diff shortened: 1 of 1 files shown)


NoahSaso (changes requested):
thanks for getting this started! could you add a more descriptive title and description to the PR? also, the goal is just to remove the USD price display, not the entire claimed balance line. We still want the label and normal amount display, just not the USD price underneath.

rishav-karanjit:
> thanks for getting this started! could you add a more descriptive title and description to the PR? also, the goal is just to remove the USD price display, not the entire claimed balance line. we still want the label and normal amount display, just not the USD price underneath.

Thank for the review. I think I misunderstood the issue. I will add a new commit for it.

packages/stateful/widgets/widgets/VestingPayments/components/VestingPaymentCard/VestingPaymentCard.tsx
```diff
@@ -495,31 +495,8 @@ export const VestingPaymentCard = ({
...
-                    />
-
-                    <TooltipInfoIcon
-                      size="xs"
-                      title={t('info.estimatedUsdValueTooltip')}
-                    />
-                  </div>
-                )}
             </div>
+
```
NoahSaso, line 499:
```suggestion
```

the linter is uspet because there's an extra line here. removing this line should satisfy it

NoahSaso (changes requested):
thank you, looks great! just one tiny adjustment needed to satisfy the linter

packages/stateful/widgets/widgets/VestingPayments/components/VestingPaymentCard/VestingPaymentCard.tsx
rishav-karanjit: Thanks. I fixed this

NoahSaso (approved):
awesome thank you 🙏

PR #1471, opened 2023-11-14
Open DAO in the same tab by default

### LINKED ISSUE #1310: Clicking a DAO should open in the same tab
Currently, clicking on a DAO in an inline link will open the link a different tab:

image

I think in-bound links should rarely if ever open in new tabs; regular navigation rules within the app should always navigate to the new page in the same tab.

If the intention here is to prevent the user from navigating away while editing, we might want to address that with a confirmation modal instead.

rishav-karanjit:

Fixes #1310 

- Currently, clicking on a DAO in an inline link will open the link a different tab. This PR will disable opening link in different tab by default but allow us to manually override it.
- This PR adds an optional openInNewTab flag to EntityDisplayProps in packages/types/stateless/EntityDisplay.tsx
- Default openInNewTab to false in the component
- Then just pass it through to the ButtonLink directly

### DIFF
--- a/packages/stateful/components/EntityDisplay.tsx
+++ b/packages/stateful/components/EntityDisplay.tsx
@@ -1,10 +1,27 @@
+import { useRouter } from 'next/router'
+
 import { EntityDisplay as StatelessEntityDisplay } from '@dao-dao/stateless'
 import { StatefulEntityDisplayProps } from '@dao-dao/types'
 
 import { useEntity } from '../hooks'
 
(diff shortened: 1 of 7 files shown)


NoahSaso (changes requested):
looks great! please add a description

rishav-karanjit:
> looks great! please add a description

Just updated the description.

NoahSaso (approved):
looks goooood

rishav-karanjit:
@NoahSaso can we close this PR to have the issue closed?

PR #1501, opened 2023-12-05
Fix USDC default and don't show in treasury by default

NoahSaso:
This hides USDC in the treasury by default since now there are multiple different USDCs and it's not very helpful. The deposit fiat buttons still show up in the treasury separately.

This also fixes the native USDC selected to prefer the pure USDC symbol (instead of USDC.axl).


JakeHartnell (approved):
:+1:

PR #1509, opened 2023-12-08
Treasury Overhaul (with refactor), ICA, Polytone expansion to Terra and Migaloo, and Skip API adoption

NoahSaso:
This is a large account system refactor that is necessary to implement Timewave's rebalancer. In the process, I also implemented a bunch of features that use the account system refactor and are needed for the rebalancer.

It:
- adds Skip API IBC routing for the Spend action and for token metadata retrieval
- adds a treasury history graph
- moves `packages/types/stateless` to `packages/types/components`
- adds ICA actions
- adds a wallet page, like the Me page, but for any wallet address
- removes WYND action, widget, and prices
- adds supported polytone connections into Terra
- adds supported polytone connections into Migaloo
- adds supported polytone connections out of Neutron
- adds membership and staking info to the home page

## Treasury History Graph

Screenshot 2023-12-14 at 12 51 42

Screenshot 2023-12-14 at 12 52 13

## IBC Spend packet-forward-middleware support

Screenshot 2023-12-13 at 17 24 06

Screenshot 2023-12-13 at 17 25 32

Screenshot 2023-12-13 at 17 24 56

Screenshot 2023-12-13 at 17 25 13

Screenshot 2023-12-13 at 17 26 31

Screenshot 2023-12-13 at 17 25 19

### Warnings when there are better paths available

Screenshot 2023-12-13 at 17 27 17

Screenshot 2023-12-13 at 17 57 32

## ICA actions

image

Screenshot 2023-12-14 at 08 56 27

### Shown on treasury page (with the same warning)

image

Screenshot 2023-12-14 at 08 57 20

## Wallet page

Screenshot 2023-12-14 at 12 51 15

## Home page membership/staking info

Screenshot 2023-12-16 at 14 40 11

### When not logged in

Screenshot 2023-12-16 at 14 40 23


JakeHartnell:
Curious what Polytone looks like on actions that support it now? I think we may want to move towards a drop down as the radio dials don't scale? Also, might look better on mobile.

THIS IS :fire: !

NoahSaso:
> Curious what Polytone looks like on actions that support it now? I think we may want to move towards a drop down as the radio dials don't scale? Also, might look better on mobile.
> 
> THIS IS 🔥 !

amazing yes good point, i will switch it to a drop down.

NoahSaso:
> Curious what Polytone looks like on actions that support it now? I think we may want to move towards a drop down as the radio dials don't scale? Also, might look better on mobile.
> 
> THIS IS 🔥 !

voila:

Screenshot 2023-12-16 at 02 19 12
Screenshot 2023-12-16 at 02 19 36

JakeHartnell:
Looks AMAZING. LFG!

JakeHartnell (approved):
This is SICK. Will make a video demo. 🚢

JakeHartnell:
Maybe if the cross-chain accounts have not been setup, we hide them? Or prompt to create an ICA first?

NoahSaso:
> 
> 
> Maybe if the cross-chain accounts have not been setup, we hide them? Or prompt to create an ICA first?

good catch. some actions, like a smart contract execute, don't require an address to be included in the message, so it's actually fine to use those because polytone will create an account for you if you don't yet have one, and then execute the message from it. some actions, like submitting a chain governance proposal, require an address in the message (proposer for a gov prop), so they need to have an account defined in order to create the message. i went ahead and made sure the actions that need premade accounts only show the chains the DAO has accounts on.

PR #1523, opened 2023-12-15
Fix AppsTab initial load when url is set

ismellike:
The URL object cannot be created when the url is encoded

### DIFF
--- a/packages/stateless/components/dao/tabs/AppsTab.tsx
+++ b/packages/stateless/components/dao/tabs/AppsTab.tsx
@@ -74,14 +74,14 @@ const InnerAppsTab = ({
   // On first iframe mount, go to url if valid already.
   useEffect(() => {
     try {
-      if (iframe && url && new URL(decodeURIComponent(url)).href) {
+      if (iframe && url && new URL(url).href) {
         iframe.src = url
       }
     } catch {
       // Ignore.
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [iframe])
+  }, [iframe, url])
 
   // If no app URL matching, choose the last one (custom) with empty URL.
   const selectedAppIndex = DAO_APPS.findIndex(
--- a/packages/stateless/components/dao/tabs/AppsTab.tsx
+++ b/packages/stateless/components/dao/tabs/AppsTab.tsx
@@ -72,7 +72,7 @@ const InnerAppsTab = ({
   // On first iframe mount, go to url
   useEffect(() => {
     try {
-      if (iframe) {
+      if (iframe && (url === '' || (url && new URL(url).href))) {
         iframe.src = url
       }
     } catch {
@@ -80,9 +80,14 @@ const InnerAppsTab = ({
     }
   }, [iframe, url])
 
+  // Update the input field to match the URL if it changes in the parent
+  // component. This should handle the URL being updated from the query params.
   useEffect(() => {
-    if(url !== inputUrl)
-      setInputUrl(url);
+    if (url !== inputUrl) {
+      setInputUrl(url)
+    }
+    // Only change the input URL when the URL changes (i.e. ignore input change).
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [setInputUrl, url])
 
   // If no app URL matching, choose the last one (custom) with empty URL.
--- a/packages/stateless/components/dao/tabs/AppsTab.tsx
+++ b/packages/stateless/components/dao/tabs/AppsTab.tsx
@@ -66,21 +66,18 @@ const InnerAppsTab = ({
 
   const go = (destUrl = url) => {
     setUrl(destUrl)
-    if (iframe) {
-      iframe.src = destUrl
-    }
   }
 
   // On first iframe mount, go to url if valid already.
   useEffect(() => {
     try {
       if (iframe && url && new URL(url).href) {
-        go(url)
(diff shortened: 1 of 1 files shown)


NoahSaso:
hey thanks for seeing an issue and taking the initiative to fix it!! i love this energy.

can you provide an example link that is broken right now but fixed by this PR? i thought the query parameters were automatically decoded by next.js 🤔

ismellike:
> hey thanks for seeing an issue and taking the initiative to fix it!! i love this energy.
> 
> can you provide an example link that is broken right now but fixed by this PR? i thought the query parameters were automatically decoded by next.js 🤔

Sure thing. You can see this by clicking any app (osmosis, stargaze, etc..) and then refreshing the page.
The following link will not automatically show the iframe: https://testnet.daodao.zone/dao/juno1rtd82r0clhjkpaye8ulpm3eqcqx53ddnsevrkung3qf064znpsws3w3lx2/apps?url=https%3A%2F%2Fapp.osmosis.zone

ismellike:
> hey thanks for seeing an issue and taking the initiative to fix it!! i love this energy.
> 
> can you provide an example link that is broken right now but fixed by this PR? i thought the query parameters were automatically decoded by next.js 🤔

Ah you're right actually. It looks like the issue is from a missing dep on url actually.

NoahSaso:
> > hey thanks for seeing an issue and taking the initiative to fix it!! i love this energy.
> > can you provide an example link that is broken right now but fixed by this PR? i thought the query parameters were automatically decoded by next.js 🤔
> 
> Ah you're right actually. It looks like the issue is from a missing dep on the url.

that was intentionally left out since the `go` function above it handles redirecting when the button is pressed... but now that i think about it, what if you tried making `go` just set the URL and let the `useEffect` handle redirecting by adding `url` back to the `useEffect` deps and removing the eslint disable line?

packages/stateless/components/dao/tabs/AppsTab.tsx
```diff
@@ -75,13 +75,12 @@ const InnerAppsTab = ({
   useEffect(() => {
     try {
       if (iframe && url && new URL(url).href) {
-        iframe.src = url
+        go(url)
```
NoahSaso, line 78: because `go` is redefined on every render (it's not wrapped in a [`useCallback`](https://react.dev/reference/react/useCallback) hook), this is actually going to fire on every render.

what i had in mind instead is that you remove the `if` statement from the `go` function so that it just contains `setUrl(destUrl)`, and then set the `useEffect` dependencies to `[iframe, url]` so that the effect hook sets `iframe.src = url` whenever the `url` changes. does that make sense? idk if it'll fix it but it's worth a shot.

NoahSaso (changes requested):

packages/stateless/components/dao/tabs/AppsTab.tsx
ismellike: Whoops there you go. Tested it and works as intended

NoahSaso:
oh shoot 🤦 now this is changing the iframe on every edit to the input. i'm not sure how often people are manually typing URLs versus copy/pasting.. but this UX isn't great and would lead to a ton of requests.

here's one solution i can think of:

have two state variables: one `inputUrl` and one `iframeUrl`. when `iframeUrl` changes, update the iframe's URL, like the current `url` variable. In `go`, set `iframeUrl` to `inputUrl`. on page load, load the current query parameter into both variables. and use `inputUrl` only for the input.

got any other ideas?

JakeHartnell (approved):

packages/stateless/components/dao/tabs/AppsTab.tsx
```diff
@@ -63,25 +63,27 @@ const InnerAppsTab = ({
...
       // Ignore.
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [iframe])
+  }, [iframe, url])
+
+  useEffect(() => {
+    if(url !== inputUrl)
+      setInputUrl(url);
+  }, [setInputUrl, url])
```
NoahSaso, line 86:
```suggestion
  // Update the input field to match the URL if it changes in the parent
  // component. This should handle the URL being updated from the query params.
  useEffect(() => {
    if (url !== inputUrl) {
      setInputUrl(url)
    }
    // Only change the input URL when the URL changes (i.e. ignore input change).
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [setInputUrl, url])
```

looks good but the formatter is going to complain. running `yarn format` before pushing commits should solve this generally.

packages/stateless/components/dao/tabs/AppsTab.tsx
```diff
@@ -63,25 +63,27 @@ const InnerAppsTab = ({
...
+  const go = (url: string) => {
+    setUrl(url)
   }
 
-  // On first iframe mount, go to url if valid already.
+  // On first iframe mount, go to url
   useEffect(() => {
     try {
-      if (iframe && url && new URL(url).href) {
+      if (iframe) {
```
NoahSaso, line 75:
```suggestion
      if (iframe && url && new URL(url).href) {
```

i think we can leave these url validation checks in

NoahSaso (changes requested):
sweet i think this logic is good! just added some formatting / clean up suggestions

NoahSaso (approved):
awesome! works perfectly, thank you 🙏

Issue #1553, opened 2023-12-24
Improve error messages


NoahSaso:
- Add denom to insufficient funds message
- Clean up proposal size too large (too many actions)

PR #1577, opened 2023-12-29
Improve some errors

NoahSaso:
The Apps tab `Unsupported chain` error was confusing users. This makes the error much more clear.

Screenshot 2023-12-29 at 00 40 08

It also adds cleaner text for proposal too large errors (part of #1553), and removes an incorrect error match.


JakeHartnell (approved):
AMAZING

PR #1623, opened 2024-01-15
Add status page to view indexer status live

NoahSaso:
This adds a status page to make checking the status of the indexers easy.

Screenshot 2024-01-14 at 18 48 14


JakeHartnell (approved):
Beautiful.

PR #1626, opened 2024-01-15
Add Neutron SubDAO pause/unpause action

NoahSaso:
This adds an action to pause and unpause Neutron SubDAOs. It is only shown for the main Neutron DAO and security SubDAO.

This also unwraps Neutron's special timelock message wrapper so that actions show correctly.

Screenshot 2024-01-15 at 14 05 44

Screenshot 2024-01-15 at 14 05 59


JakeHartnell (approved):
LGTM

Issue #1599, opened 2024-01-08
Display NFT floor prices and count in TVL


NoahSaso:
Since we now support DAOs and cross-chain accounts on Stargaze, DAOs are beginning to hold valuable assets. NFTs are currently not included in TVL, nor are their floor prices displayed.

We should load prices from:
- Stargaze
- Loop on Juno

NoahSaso:
`packages/types/nft.ts` has `NftCardInfo` which already has a `floorPrice` field:
```
  floorPrice?: {
    amount: number
    denom: string
  }
```

it's never been used, and you may want to replace `denom` with the `GenericToken` type which is now used more widely throughout the UI

the NFTs a DAO has are loaded in `lazyNftCardInfosForDaoSelector` in `packages/stateful/recoil/selectors/nft.ts`, but only the minimal info needed to pass to cards that load the actual data later. the actual NFT info is loaded in `nftCardInfoWithUriSelector`.

adding it to DAO TVL is going to be trickier. that's done in `packages/state/recoil/selectors/treasury.ts` in `daoTvlSelector`. right now it's only setup for tokens. a potential solution is to actually add the floor price to the `LazyNftCardInfo` type so you can retrieve it in `lazyNftCardInfosForDaoSelector`, and then use that in the TVL selector.

NoahSaso:
to load the floor price from stargaze, you'll probably want to add a new stargaze graphql query in `packages/state/graphql/stargaze/queries/collection.ts`

the graphql playground is here: https://graphql.mainnet.stargaze-apis.com/graphql
and you can retrieve the `floorPrice` from the `collection` object.

after adding a graphql query, you'll need to run `yarn build:gql` in the state package (or `yarn state build:gql` from the root) to rebuild the graphql typescript types.

you can see an example of how graphql queries are accessed in `packages/state/recoil/selectors/contracts/CommonNft.ts` in `unpaginatedAllTokensForOwnerSelector`

NoahSaso:
Closed by #1651 

thanks @ismellike !!

PR #1651, opened 2024-01-21
Display stargaze NFT's highest offer and include NFT value in DAO TVL

ismellike:
#1599 - Only doing Stargaze for now, because I could not get more info from Loop. 
The new  `allNftUsdValueSelector` in CommonNft is there to plug the data in whenever accessible though.


packages/state/recoil/selectors/contracts/CommonNft.ts
```diff
@@ -45,6 +47,38 @@ export const queryClient = selectorFamily<
...
+              })
+            )
+            break
+
+          default:
+            break
+        }
+      })
+
+      return sum
```
NoahSaso, line 78: a couple things:

first, this `CommonNft` file is for standardizing smart contract queries between cw721 and sg721 contracts, so this selector should probably be in `packages/state/recoil/selectors/nft.ts`. after moving this, you might want to change the `key` too.

also, recoil has this awesome [`waitForAll`](https://recoiljs.org/docs/api-reference/utils/waitForAll/) utility that could parallelize this request:

```suggestion
      const sum = get(
        waitForAll(
          accounts
            .filter(
              (x) =>
                x.chainId === ChainId.StargazeMainnet ||
                x.chainId === ChainId.StargazeTestnet
            )
            .map((x) =>
              stargazeWalletUsdValueSelector({
                chainId: x.chainId,
                address: x.address,
              })
            )
        )
      ).reduce((acc, x) => acc + x, 0)

      return sum
```

packages/state/recoil/selectors/stargaze.ts
```diff
@@ -0,0 +1,44 @@
...
+import {
+  stargazeIndexerClient,
+  stargazeWalletTotalValueUsdQuery,
+} from '../../graphql'
+
+export const stargazeWalletUsdValueSelector = selectorFamily<
+  number,
+  WithChainId<{ address: string }>
+>({
+  key: 'commonNftStargazeWalletUsdValue',
```
NoahSaso, line 14:
```suggestion
  key: 'stargazeWalletUsdValue',
```

packages/state/recoil/selectors/stargaze.ts
```diff
@@ -0,0 +1,44 @@
...
+  get:
+    ({ chainId, address }) =>
+    async () => {
+      if (
+        chainId !== ChainId.StargazeMainnet &&
+        chainId !== ChainId.StargazeTestnet
+      ) {
+        throw new Error('Expected Stargaze mainnet chain')
```
NoahSaso, line 22:
```suggestion
        throw new Error('Expected Stargaze chain')
```

packages/state/recoil/selectors/stargaze.ts
```diff
@@ -0,0 +1,44 @@
...
+
+      if (error) {
+        throw error
+      }
+
+      if (
+        data.wallet?.stats?.totalValueUsd === null ||
+        data.wallet?.stats?.totalValueUsd === undefined
+      )
+        throw new Error('Unexpected response from Stargaze indexer')
```
NoahSaso, line 40: it's possible stargaze will error like this if the wallet hasn't been seen before, so i think it would be safer to return 0 if the data seems malformed. also `data` may be undefined, so this should probably check for that too

```suggestion
      if (!data?.wallet?.stats?.totalValueUsd) {
        return 0
      }
```

packages/utils/nft.ts
```diff
@@ -88,4 +90,18 @@ export const nftCardInfoFromStargazeIndexerNft = (
...
+          amount:
+            token.highestOffer.offerPrice.amount &&
+            convertMicroDenomToDenomWithDecimals(
+              token.highestOffer.offerPrice.amount,
+              6
+            ),
+          amountUsd: token.highestOffer.offerPrice.amountUsd,
+          denom: 'STARS',
+        }
+      : undefined,
```
NoahSaso, line 105: `offerPrice` seems to support different denoms, so we can't assume that this is a STARS token with 6 decimals. instead, i think this helper function should take a `token: GenericToken` argument, which you can retrieve in the selector that uses this helper function with `get(genericTokenSelector(...))`

also instead of doing any conversion here, `highestOffer` should contain that `token` field directly so it can be referenced wherever necessary, and perform the amount decimal conversions right in the component.

packages/stateless/components/NftCard.tsx
```diff
@@ -262,16 +264,29 @@ export const NftCard = forwardRef<HTMLDivElement, NftCardProps>(
...
-                <p className="secondary-text">{t('title.floorPrice')}</p>
-
+                <p className="secondary-text">{t('title.highestOffer')}</p>
                 <p className="body-text font-mono">
-                  {floorPrice.amount.toLocaleString(undefined, {
+                  {highestOffer.amount?.toLocaleString(undefined, {
                     maximumSignificantDigits: 3,
                   })}{' '}
-                  ${floorPrice.denom}
+                  ${highestOffer.denom}
```
NoahSaso, line 275: can you replace this with a `TokenAmountDisplay` component like the USD value below? that way it will render a token icon and look consistent. see my other comment about storing the token metadata directly in `highestOffer` so you can access its information here.

packages/types/nft.ts
```diff
@@ -63,16 +70,18 @@ export type NftCardInfo = {
   imageUrl?: string
   // Metadata loaded from the token URI.
   metadata?: Record<string, any>
-  floorPrice?: {
-    amount: number
-    denom: string
+  highestOffer?: {
+    amount?: number | null
+    amountUsd?: number | null
+    denom?: string | null
```
NoahSaso, line 76: if there's no highest offer, can we fallback to collection floor price?

NoahSaso (changes requested):
great start! you really found your way around the codebase 😁🙌

packages/types/nft.ts
ismellike: Stargaze's API handles this already. It looks like they search through 4 sources

NAME_OFFER
TOKEN_OFFER
COLLECTION_OFFER
LIVE_AUCTION_OFFER

NoahSaso: oh amazing! that makes life so much easier for us

packages/state/recoil/selectors/nft.ts
```diff
@@ -76,6 +79,35 @@ export const nftUriDataSelector = selectorFamily<
   },
 })
 
+export const allNftUsdValueSelector = selectorFamily<
+  number,
+  WithChainId<{ address: string }>
+>({
+  key: 'commonNftAllNftUsdValue',
```
NoahSaso, line 86:
```suggestion
  key: 'allNftUsdValue',
```

packages/state/recoil/selectors/nft.ts
```diff
@@ -76,6 +79,35 @@ export const nftUriDataSelector = selectorFamily<
...
+            .filter(
+              (x) =>
+                x.chainId === ChainId.StargazeMainnet ||
+                x.chainId === ChainId.StargazeTestnet
+            )
+            .map((x) =>
+              stargazeWalletUsdValueSelector({
+                chainId: x.chainId,
+                address: x.address,
+              })
```
NoahSaso, line 103:
```suggestion
            .filter(
              ({ chainId }) =>
                chainId === ChainId.StargazeMainnet ||
                chainId === ChainId.StargazeTestnet
            )
            .map(({ chainId, address }) =>
              stargazeWalletUsdValueSelector({
                chainId,
                address,
              })
```

style nit: i prefer immediately destructuring fields like this, which makes the selector params object cleaner and gets rid of a meaningless variable (`x`). not a big deal and i don't expect you to perfectly follow my style preferences, but i wanted to let you know that's an option :)

packages/stateful/recoil/selectors/nft.ts
```diff
@@ -66,12 +74,38 @@ export const walletStargazeNftCardInfosSelector = selectorFamily<
...
+            console.log(token)
+            return nftCardInfoFromStargazeIndexerNft(
+              chainId,
+              token,
+              token.highestOffer?.offerPrice?.denom
+                ? genericTokensMap.get(token.highestOffer.offerPrice.denom)
+                : undefined,
+              timestamp
+            )
+          })
```
NoahSaso, line 106:
```suggestion
          ...data.tokens.tokens.map((token) =>
            nftCardInfoFromStargazeIndexerNft(
              chainId,
              token,
              token.highestOffer?.offerPrice?.denom
                ? genericTokensMap.get(token.highestOffer.offerPrice.denom)
                : undefined,
              timestamp
            )
          )
```

extraneous log

packages/stateless/components/NftCard.tsx
```diff
@@ -262,17 +265,37 @@ export const NftCard = forwardRef<HTMLDivElement, NftCardProps>(
...
 
-            {/* Floor price */}
-            {floorPrice && (
+            {/* Highest offer */}
+            {highestOffer && (
               <div className="space-y-2">
-                <p className="secondary-text">{t('title.floorPrice')}</p>
+                <p className="secondary-text">{t('title.highestOffer')}</p>
+                <div className="body-text font-mono">
+                  {highestOffer.amount && highestOffer.offerToken && (
```
NoahSaso, line 273:
```suggestion
                  {typeof highestOffer.amount === 'number' && highestOffer.offerToken && (
```

if amount is `0`, it is falsey but react will still render a `0`, so an explicit numeric check is better here

packages/stateless/components/NftCard.tsx
```diff
@@ -262,17 +265,37 @@ export const NftCard = forwardRef<HTMLDivElement, NftCardProps>(
...
             )}
 
-            {/* Floor price */}
-            {floorPrice && (
+            {/* Highest offer */}
+            {highestOffer && (
               <div className="space-y-2">
-                <p className="secondary-text">{t('title.floorPrice')}</p>
+                <p className="secondary-text">{t('title.highestOffer')}</p>
+                <div className="body-text font-mono">
```
NoahSaso, line 272: is there supposed to be `space-y-1` in this div or no? just curious because i don't see any other margins

packages/utils/nft.ts
```diff
@@ -88,4 +90,12 @@ export const nftCardInfoFromStargazeIndexerNft = (
   imageUrl: token.media?.visualAssets?.lg?.url || token.media?.url || undefined,
   name: token.name || token.tokenId || 'Unknown NFT',
   description: token.description || undefined,
+  highestOffer: offerToken
+    ? {
+        offerToken: offerToken,
```
NoahSaso, line 95:
```suggestion
        offerToken,
```

style nit: if you want

NoahSaso (changes requested):
you're officially goated.. nice

i made a couple of style nits, for your consideration; only one or two of my comments actually require fixing

NoahSaso:
it seems like the USD value is being shown but not the STARS price:

https://dapp-mainnet-git-fork-ismellike-ismellike-nft-tvl-fbbec3-da0da0.vercel.app/dao/juno1q6v7qhq0sepf87ra8gnwcmmj3r6em5vqucrea932tjwhvqzdm2fs0qsysf/treasury

Screenshot 2024-01-24 at 16 17 40

NoahSaso:
Screenshot 2024-01-25 at 09 43 41

ok it works! i think it's a bit cramped now

adding `ml-7` to the price container to make it align with the token amount looks like this, i think it's a bit better:
Screenshot 2024-01-25 at 09 45 53

wdyt?

ismellike:
> Screenshot 2024-01-25 at 09 43 41
> ok it works! i think it's a bit cramped now
> 
> adding `ml-7` to the price container to make it align with the token amount looks like this, i think it's a bit better: Screenshot 2024-01-25 at 09 45 53
> 
> wdyt?

Yep I like that too

packages/stateless/components/NftCard.tsx
```diff
@@ -262,17 +265,43 @@ export const NftCard = forwardRef<HTMLDivElement, NftCardProps>(
...
+                        decimals={highestOffer.offerToken.decimals}
+                        iconUrl={highestOffer.offerToken.imageUrl}
+                        symbol={highestOffer.offerToken.symbol}
+                      />
+                    )}
+                  {highestOffer.amountUsd && (
+                    <div
+                      className={`caption-text flex flex-row items-center gap-1${
+                        highestOffer.offerToken?.imageUrl ? ' ml-7' : ''
+                      }`}
```
NoahSaso, line 290: we use `clsx` to clean up this common variable class name issue

```suggestion
                      className={clsx(
                        'caption-text flex flex-row items-center gap-1',
                        !!highestOffer.offerToken?.imageUrl && 'ml-7'
                      )}
```

it basically just ignores any falsey argument and combines string arguments into a single string

NoahSaso (commented):
one more style nit <3

packages/stateless/components/NftCard.tsx
ismellike: Cool! Learning a lot from this

NoahSaso (approved):
awesome!! thank you for tackling this 🤝

Issue #1172, opened 2023-03-10
Widget for Social links


JakeHartnell:
DAOs want to display social links: https://twitter.com/RAC_FM_Spaces/status/1634013454100213760?s=20

Could we support this with a widget?

JakeHartnell:
What's blocking this?

NoahSaso:
> What's blocking this?

design

pageofswrds:
How do I assign this to myself

Issue #1526, opened 2023-12-16
Add a way to view all active authorizations


NoahSaso:

JakeHartnell:
Awesome, this could be a new widget.

Issue #1525, opened 2023-12-16
Add a way to view all contracts a DAO is the admin of


NoahSaso:
Requested here: https://x.com/kais_kat/status/1736155041793204499?s=20

JakeHartnell:
This would be nice, in addition it would be nice to see all of the contracts the DAO is comprised of. Sometimes, you still need to find those addresses for custom props.

PR #1678, opened 2024-01-29
Redesign

### LINKED ISSUE #1223: Followed SubDAOs should only show up once.
image

People will often follow SubDAOs. That leads to a situation where SubDAOs show up twice in the sidebar (shown above).

SubDAOs you're following should only show up once. 

My suggestion:

- If you follow a SubDAO and its parent DAO, place a ["following" icon](https://thenounproject.com/browse/icons/term/following/) next to SubDAOs you follow.

- If you follow a SubDAO and *not* its parent, show the SubDAO with a "following" icon next to it, and *only* the top-level DAO as its parent (but no sibling DAOs).

Curious to hear any other suggestions to deal with this situation.

### LINKED ISSUE #1243: Not clear you cannot vote on props created before you stake
If you staked _after_ a proposal was created, you can't vote on that proposal. We should make this crystal clear, a bunch of folks voiced frustration / confusion about this during the DAO FUD spaces.

### LINKED ISSUE #1244: Mobile UX: not clear where the wallet is
14C84018-CA65-423A-A6F9-8A2E9D4359EB

During the DAO FUD spaces, some people pointed out that when the wallet is connected, it’s not very clear how to get to it. The circle is a little too hidden, and if a new user hasn’t added a pic to their profile it’s even less clear.

Would be good to compare mobile UX here to other dapps.

### LINKED ISSUE #1489: No way to navigate to chain-specific home page
The home page that shows feed, featured, and followed DAOs shows DAOs on all chains. Each supported chain also has its own page (e.g. `/stargaze`) that only shows featured and followed DAOs from that chain, but there's no way to navigate to these pages without knowing they exist and manually typing them into the browser.

One way to solve this would be to add a chain switcher to the home page with an extra `All Chains` option that is selected by default on the main home page.

### LINKED ISSUE #1680: Truncate unknown factory denoms
This is ungodly.

Also, for many LP factory tokens, we should be able to query the contract and fetch info about the pair.

telegram-cloud-photo-size-2-5314559199285467690-y

NoahSaso:
Resolves #1223 
Resolves #1489 
Resolves #1244
Resolves #1680
Resolves #1243

Long overdue redesign. This basically makes DAO DAO mobile-native.

I'll try to list everything done:
- abolishes the right sidebar
- adds a chain switcher to the home page
- mobile dock (bottom nav)
- wallet in top right on desktop
- wallet in dock and nav sidebar on mobile
- treasury lines instead of cards that click to open the card modal
- more modern DAO page with horizontally scrolling tab bar
- home page with all DAO config cards shown
- proposal creation page shows all relevant config
- renamed Inbox to Notifications
- renamed Me to Account
- renamed Governance to Chains
- replaced many loaders with placeholder cards
- more

It also cleans up tons of random stuff.


JakeHartnell:
This is a huge step forward!

**Few notes from my first pass...**

1. Many widgets leads to hidden widgets... not super obvious that you have to scroll sideways to get to the "Vesting Payments" widget here:
image
image

This is even more pronounced on mobile, it's not obvious to new users to side scroll.
image

2. If a settings card is "N / A" we should just not show it to reduce mental overload on the user.
image

3. Don’t show not installed wallets on mobile. Hides social logins.
image

4. Make it more obvious that a long DAO description is truncated, a simple "read more" would do the trick.
image

**Some bonus things that may make sense to do now:**

1. A widget for social links would look great on this new home page. Would close #1172 and be very useful IMO, could be simple... also fine to do separately if it's too much work for now.
image

2. On mobile, tool tips disappear too quickly to read them. Visit the site on mobile, touch one of the question marks and observe the behavior (can file a separate issue for this if need be as it existed before the redesign).

3. Is there no more SDA now? We do have a few people that use the dedicated DAO pages and the PMF for this came up recently in user testing with someone who wants to use DAO DAO for a project and have members of that project not get distracted by other DAOs. They really appreciated the SDA, but I realize that's more to redesign and could come later.

4. Would be nice to see all of the contracts the DAO is comprised of and other advanced info such as authorizations. Related to #1525 and #1526. Perhaps this also belongs on the home page? Could be done later.

CLHRAE (approved):

PR #1684, opened 2024-02-04
Add auth/deauth voter actions

noahsaso:
<3 @ShayneRarma


noahsaso:
closing this as we are building real delegation into the contracts

PR #1694, opened 2024-02-08
Nft unstaking duration

ismellike:
How does this look? @NoahSaso @CLHRAE
I feel like the NFT selection is starting to get a little cramped. The container should at least show the full NFT card.
Any suggestions? 

image


noahsaso:
> How does this look? @NoahSaso @CLHRAE I feel like the NFT selection is starting to get a little cramped. The container should at least show the full NFT card. Any suggestions?
> 
> image


yeah agreed, this feels too cramped. what if you left everything the same as before, and instead just added the `It will take 2 weeks...` right under the `Stake NFTs`/`Unstake NFTs` buttons? it should probably show up all the time.

ismellike:
> > How does this look? @NoahSaso @CLHRAE I feel like the NFT selection is starting to get a little cramped. The container should at least show the full NFT card. Any suggestions?
> > image
> 
> yeah agreed, this feels too cramped. what if you left everything the same as before, and instead just added the `It will take 2 weeks...` right under the `Stake NFTs`/`Unstake NFTs` buttons? it should probably show up all the time.

I found a typical usage of the `TooltipInfoIcon` in `DaoVotingVaultCard` and applied it at the new location. Here's how it looks now:

image

noahsaso:
> > > How does this look? @NoahSaso @CLHRAE I feel like the NFT selection is starting to get a little cramped. The container should at least show the full NFT card. Any suggestions?
> > > image
> > 
> > 
> > yeah agreed, this feels too cramped. what if you left everything the same as before, and instead just added the `It will take 2 weeks...` right under the `Stake NFTs`/`Unstake NFTs` buttons? it should probably show up all the time.
> 
> I found a typical usage of the `TooltipInfoIcon` in `DaoVotingVaultCard` and applied it at the new location. Here's how it looks now:
> 
> image

oh sir that's brilliant! what do you think about moving it above the stake/unstake buttons? so it's under the subtitle.

packages/types/nft.ts
```diff
@@ -1,6 +1,6 @@
 import { ComponentType, ReactNode, RefAttributes } from 'react'
 
-import { GenericToken } from '.'
+import { Duration, GenericToken } from '.'
```
noahsaso: can you change this import to not import from the index? ie import `GenericToken` from `./token` and `Duration` from its own file too

packages/stateful/components/nft/NftSelectionModal.tsx
```diff
@@ -203,6 +209,30 @@ export const NftSelectionModal = ({
         nfts.errored ||
         nfts.data.length > 0 ? (
           <div className="mt-4 flex flex-col gap-4">
+            {unstakingDuration &&
+              ('height' in unstakingDuration
+                ? unstakingDuration.height
+                : unstakingDuration.time) > 0 && (
+                <div className="flex flex-row items-center gap-1">
```
noahsaso:
```suggestion
                <div className="-mt-3 flex flex-row items-center gap-1">
```

there's a bit too much spacing now, let's remove some

noahsaso (changes requested):
awesome! i think this is the cleanest we'll get for now. thanks :)

just two tiny changes

noahsaso (approved):
yay! thank you

Issue #1683, opened 2024-02-03
Show tokens DAOs staked in other DAOs in treasury


NoahSaso:
We should detect when DAOs are staking in other DAOs and show it. I think we do this for wallets, but I guess not DAOs.

PR #1714, opened 2024-02-19
Detect native denoms fully staked in daos

ismellike:
#1683 
I also added an eslint rule to fix this issue for windows users:
https://stackoverflow.com/questions/53516594/why-do-i-keep-getting-eslint-delete-cr-prettier-prettier

### DIFF
--- a/packages/state/recoil/selectors/chain.ts
+++ b/packages/state/recoil/selectors/chain.ts
@@ -3,6 +3,7 @@ import { fromBase64, toHex } from '@cosmjs/encoding'
 import { Coin, IndexedTx, StargateClient } from '@cosmjs/stargate'
 import uniq from 'lodash.uniq'
 import {
+  noWait,
   selector,
   selectorFamily,
   waitForAll,
@@ -78,7 +79,7 @@ import {
   queryValidatorIndexerSelector,
 } from './indexer'
 import { genericTokenSelector } from './token'
-import { walletTokenStakedDenomsSelector } from './wallet'
+import { walletTokenDaoStakedDenomsSelector } from './wallet'
 
 export const stargateClientForChainSelector = selectorFamily<
   StargateClient,
@@ -344,9 +345,15 @@ export const nativeBalancesSelector = selectorFamily<
         ...get(justNativeBalancesSelector({ address, chainId })),
       ]
       const nativeToken = getNativeTokenForChainId(chainId)
-      const stakedDenoms = get(
-        walletTokenStakedDenomsSelector({ walletAddress: address, chainId })
-      )
+      const stakedDenoms =
+        get(
+          noWait(
+            walletTokenDaoStakedDenomsSelector({
+              walletAddress: address,
+              chainId,
+            })
+          )
+        ).valueMaybe() || []
 
       const uniqueDenoms = new Set(balances.map(({ denom }) => denom))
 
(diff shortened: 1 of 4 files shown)


packages/state/recoil/selectors/wallet.ts
```diff
@@ -50,3 +54,73 @@ export const walletCw20BalancesSelector = selectorFamily<
       }))
     },
 })
+
+export const walletTokenStakedDenomsSelector = selectorFamily<
+  readonly string[],
+  WithChainId<{ walletAddress: string }>
+>({
+  key: 'walletTokenStakedDenoms',
```
noahsaso:
```suggestion
export const walletTokenDaoStakedDenomsSelector = selectorFamily<
  readonly string[],
  WithChainId<{ walletAddress: string }>
>({
  key: 'walletTokenDaoStakedDenoms',
```

can we rename this to make it a bit more clear the kinds of staked tokens?

packages/state/recoil/selectors/wallet.ts
```diff
@@ -50,3 +54,73 @@ export const walletCw20BalancesSelector = selectorFamily<
...
+  get:
+    ({ walletAddress, chainId }) =>
+    ({ get }) => {
+      // Get the DAO's that the wallet is a member of
```
noahsaso:
```suggestion
      // Get the DAOs that the wallet is a member of
```

typo

packages/state/recoil/selectors/chain.ts
```diff
@@ -342,15 +343,30 @@ export const nativeBalancesSelector = selectorFamily<
...
+      // Add native denom if not present.
+      if (!uniqueDenoms.has(nativeToken.denomOrAddress)) {
         balances.push({
           amount: '0',
           denom: nativeToken.denomOrAddress,
         })
+        uniqueDenoms.add(nativeToken.denomOrAddress)
       }
 
+      // Add denoms staked to dao's if not present.
```
noahsaso:
```suggestion
      // Add denoms staked to DAOs if not present.
```

typo

packages/state/recoil/selectors/chain.ts
```diff
@@ -342,15 +343,30 @@ export const nativeBalancesSelector = selectorFamily<
       const balances = [
         ...get(justNativeBalancesSelector({ address, chainId })),
       ]
-      // Add native denom if not present.
       const nativeToken = getNativeTokenForChainId(chainId)
-      if (!balances.some(({ denom }) => denom === nativeToken.denomOrAddress)) {
+      const stakedDenoms = get(
+        walletTokenStakedDenomsSelector({ walletAddress: address, chainId })
+      )
```
noahsaso:
```suggestion
      const stakedDenoms = get(
        noWait(
          walletTokenStakedDenomsSelector({ walletAddress: address, chainId })
        )
      ).valueMaybe() || []
```

this query might take extra time depending on the number of DAOs the user interacts with, so let's make it non-blocking! `noWait` is a handy utility recoil function to do this :)

`noWait` turns it into a loadable, and then `valueMaybe` returns the value once it's loaded, and undefined otherwise. when not loaded, this just defaults to an empty array so the rest of the logic can remain the same

packages/state/recoil/selectors/wallet.ts
```diff
@@ -50,3 +54,73 @@ export const walletCw20BalancesSelector = selectorFamily<
...
+            names: DAO_VOTING_TOKEN_STAKED_CONTRACT_NAMES,
+          })
+        )
+      )
+
+      if (
+        !votingModules ||
+        !Array.isArray(votingModules) ||
+        votingModules.length === 0
+      ) {
```
noahsaso:
```suggestion
      if (votingModules.length === 0) {
```

no need for these conditions as `votingModules` is always defined and always an array.

the reason we need to check `daos` above in this way is because it's returned from an external API (our indexer), and we want to safeguard against any unexpected responses.

noahsaso (changes requested):
awesome work!! some tiny nits but no functionality changes :)

noahsaso (approved):
thank you!

PR #1736, opened 2024-03-01
Fix cw20 vesting creation by using a string amount

ismellike:
image

Fixes this error, because cw20 messages are sent as a number instead of as a string (Uint128)


noahsaso:
🙏

Issue #1723, opened 2024-02-21
Show staked assets, that are staked to other DAO's


mOdeToken:
In the Treasury of every DAO you are able to see all the assets you manage and also where are they staked. Is it possible to make it so that you could also see assets that are staked on other DAOs?

For example we have a separate DAO structure - Nearchos, that holds some $KLEO tokens. We have staked those tokens on Kleomedes DAO and now the treasury simply says that we own 0 $KLEO, even tho they are staked on DAO DAO.

noahsaso:
that's a bug, it should be showing your $KLEO stake :0
thanks for submitting this!

noahsaso:
fixed!

this was actually due to Kleomedes using a different code ID of our cw20 staking voting module: `1495`. how dare they 😠

PR #1746, opened 2024-03-07
Progress on fixing staked cw20

ismellike:
#1723 

daos/memberOf is returning empty with the cw20 dao atm, but I think the logic makes sense


ismellike:
Fixed via indexer by noah

PR #1749, opened 2024-03-07
Dynamically load react-codemirror2

ismellike:
Fixes error when trying to run development instance: Error [ReferenceError]: document is not defined https://nextjs.org/docs/pages/building-your-application/optimizing/lazy-loading#with-no-ssr


noahsaso (approved):
thanks!

PR #1775, opened 2024-05-17
Add Galaxy Station wallet

echel0n-HX:


noahsaso:
hey thanks for this! is there a version of `@cosmos-kit/galaxy-station` that depends on `@cosmos-kit/core` version `2.10.0`? would rather not install multiple versions of the same package, and it may cause type issues

echel0n-HX:
Good catch, I've submitted a PR to cosmos-kit to bump the version, once that happens I will then update this PR with the changes.

echel0n-HX:
As requested, its been upgraded to use cosmos-kit v2.10.0

noahsaso:
> As requested, its been upgraded to use cosmos-kit v2.10.0

actually now it looks to be using `v2.11.0` 😅 i'll push a commit updating the rest of the packages

noahsaso (approved):
thanks!

PR #1779, opened 2024-05-29
Become SubDAO

nikitatrifan:
Introducing the two new actions, letting DAOs become SubDAOs.
- Become SubDAO to nominate its admin (This is done on the SubDAO side)
- Approve SubDAO join to accept the admin nomination (This is done on the Parent DAO side)

image


packages/stateful/actions/core/dao_governance/BecomeSubDao/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+            msg: {
+              nominate_admin: {
+                admin,
+              },
+            },
+          },
+        },
+      })
+  }
+
```
nikitatrifan: @NoahSaso will need to add a little validation here, see if this dao is already a subdao and if that's the case this action won't be available. Trying to think, what do you think is the best play here? Are there any queries we can shoot to validate for it

noahsaso: yep! you can get `context` from the options parameter of the `makeBecomeSubDaoAction` function, and then in the return, add:

```ts
// If parent DAO exists, hide this action.
hideFromPicker: context.type === ActionContextType.Dao && context.info.parentDao !== null
```

packages/stateful/actions/core/dao_governance/BecomeSubDao/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+    return {
+      match: true,
+      data: {
+        admin: msg.wasm.execute.nominate_admin.admin,
+      },
+    }
+  } catch (e) {
+    return { match: false }
+  }
+}
```
noahsaso: this is such a savage way of doing this, wow. i love it.

packages/stateful/actions/core/dao_governance/BecomeSubDao/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+  msg: Record<string, any>
+) => {
+  try {
+    return {
+      match: true,
+      data: {
+        admin: msg.wasm.execute.nominate_admin.admin,
+      },
+    }
+  } catch (e) {
```
noahsaso:
```suggestion
  } catch {
```

drop the unused variable

packages/stateful/actions/core/dao_governance/AcceptSubDao/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  msg: Record<string, any>
+) => {
+  try {
+    return {
+      match: true,
+      data: {
+        address: msg.wasm.execute.contract_addr,
+      },
+    }
+  } catch (e) {
```
noahsaso:
```suggestion
  } catch {
```

drop the unused variable

packages/stateful/actions/core/dao_governance/AcceptSubDao/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return { match: false }
+  }
+}
```
noahsaso: so this is going to match any execute smart contract action, because you never check to make sure the `msg` is for this specific execute variant.

i recommend using the `objectMatchesStructure` util (take a look at other actions that use this in the decoder)

packages/stateful/actions/core/dao_governance/AcceptSubDao/README.md
```diff
@@ -0,0 +1,20 @@
...
+Accept SubDAO join request and complete its ownership transfer.
+
+## Bulk import format
+
+This is relevant when bulk importing actions, as described in [this
+guide](https://github.com/DA0-DA0/dao-dao-ui/wiki/Bulk-importing-actions).
+
+### Key
+
+`becomeSubDao`
```
noahsaso:
```suggestion
`acceptSubDao`
```

noahsaso (changes requested):
love how simple this is, nice :)

just a few small fixes.

also something we need to figure out: after the parent accepts the admin nomination, it also needs to execute on behalf of the SubDAO (with its newly obtained admin privileges) the message to update the wasm contract-level admin to the parent. this is required so that the SubDAO can't just upgrade its own contract later and take power back from the parent, and also so the parent obtains the permission to upgrade it instead. to do this, in the parent's accept subDAO action, we can programmatically add the update admin action

packages/stateful/actions/core/dao_governance/AcceptSubDao/index.tsx
nikitatrifan: Good god, good catch man

packages/stateful/actions/core/dao_governance/AcceptSubDao/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const useDefaults: UseDefaults<AcceptSubDaoData> = () => defaultAcceptSubDaoData
+
+const useDecodedCosmosMsg: UseDecodedCosmosMsg<AcceptSubDaoData> = (
+  msg: Record<string, any>
+) => {
+  try {
+    const match = Boolean(
+      msg.wasm.execute.contract_addr.msg.accept_admin_nomination
+    )
```
noahsaso:
```suggestion
    const match = Boolean(msg.wasm.execute.msg.accept_admin_nomination)
```

extraneous `contract_addr`

noahsaso (changes requested):
noice :) one more thing

noahsaso (approved):
thanks for getting this setup @nikitatrifan ! 🙏

PR #1783, opened 2024-06-05
React Query!

noahsaso:
Start adopting React Query to improve performance, with the goal of completely removing Recoil once all queries are migrated.


elix1er:
lovely to hear

PR #1794, opened 2024-06-26
Sort proposal modules by prefix before selection

ismellike:
In the Arena DAO, we have 2 single proposal modules with different prefixes 'A' and 'B'. We want 'A' to be preferred, because 'B' does not allow all types of messages to be proposed. If we do a simple sort by prefix, then this is fixed.

image


noahsaso:
ah nice! i remember you mentioned this a while ago, and i thought i fixed it, but i must have just fixed the order in the selector and not actually the default. thanks for following up with this!

what if instead of sorting the array where it's used, you just sort the source proposal modules list (in the DAO info retrieval query)?

ismellike:
yes, much better now

noahsaso (approved):
awesome 🙏 thank you!

PR #1802, opened 2024-07-09
Fix funds not being sent in prepropose.propose

ismellike:
Native proposal deposits were erroring out due to the funds not being sent.

`Query failed with (6): rpc error: code = Unknown desc = failed to execute message; message index: 0: No funds sent: execute wasm contract failed`


noahsaso (approved):
THANK YOU

PR #1821, opened 2024-10-15
New Feat: Dockerize apps

hard-nett:
Docker images can be built with:
- sda image: `SDA_IMAGE=da0-da0/dao-app-sda:v0.0.2`
- dapp image: `DAPP_IMAGE=da0-da0/dao-app-dapp:v0.0.2`

*Compiling the images seem to require at minimum 16GB RAM, & 50GB of storage.*

on ARM platform:
`docker-compose build --build-arg BUILDPLATFORM=linux/arm64 --build-arg TARGETPLATFORM=linux/amd64`

on AMD platform: 
`docker-compose build --build-arg BUILDPLATFORM=linux/amd64 --build-arg TARGETPLATFORM=linux/amd64`

## Akash SDL
There is also an example of an SDL file that can be referenced for deploying both app images (dapp & sda) on machines rented via the permissioneess compute market on Akash.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+node_modules
+package-lock.json
+.git
+.gitignore
+.vscode
+.next
+.turbo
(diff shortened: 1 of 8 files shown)


noahsaso (approved):
awesome!! thank you

PR #1832, opened 2025-02-01
docs: clarify chainId requirement for AuthzExec messages

FoxinouFR:


noahsaso:
actually this `chainId` parameter is always required, so i just went ahead and added it to the README: https://github.com/DA0-DA0/dao-dao-ui/commit/89b81e38834b93385cdd9815b9a9481cdd19b86d

thanks for pointing this out!

PR #1846, opened 2026-03-30
Add a contract_info query fallback for newer cw721 contract variants

MightOfOaks:
A small change to add a fallback to cover new cw721 contract variants in case the initial `contract_info` query fails. 

Newer cw721 contracts created for migrating collections from Stargaze to Cosmos Hub do not support the `contract_info` query (marked as deprecated on cw721-base v0.19 and on) and replace it with a `get_collection_info_and_extension` query instead. The change results in an error when trying to create an NFT-based DAO. The `get_collection_info_and_extension` response extends the `contract_info` response and includes the `name` and `symbol` information needed by the DAODAO UI.


noahsaso (approved):

noahsaso:
thanks!

